PR #2, opened 2018-11-12
PlainDB for multi-trie multi-hash

sorpaas:
This lays out the traits needed to solve https://github.com/paritytech/substrate/issues/872#issuecomment-436816629, to support multiple hashing algorithms in multi-trie.

Because we can't know how many children tries are created, multi-trie needs to write all hashdb values into a single storage. To support multiple hash algorithms, the length of the hash might also be different, while the underlying data store might expect a fixed length hash. With traits in this PR, we can fix both of the issues:

* Pass `PlainDB` reference in children delta trie root generation. The generation function wraps `PlainDB` into another `HashDB` of the targeted hash algorithm to continue the operation.
* If the generation function is really sure that targeted hash and original hash have the same length (and hashes won't collide), then the wrapper can directly use targeted hash as original hash and pass it back to `PlainDB`.
* Otherwise, the wrapper can use original hash algorithm to do `H::hash(key)` before passing it back to `PlainDB`.

This PR should be backward-compatible. The only case where it breaks is if a file tries to do `use hash_db::*`.

### DIFF
--- a/hash-db/Cargo.toml
+++ b/hash-db/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "hash-db"
-version = "0.9.0"
+version = "0.11.0"
 authors = ["Parity Technologies <admin@parity.io>"]
 description = "Trait for hash-keyed databases."
 license = "Apache-2.0"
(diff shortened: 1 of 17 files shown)


sorpaas:
Figured out that `HashDB::keys` won't work with the above design. However, this function doesn't seem to be used anywhere. So I moved it to `impl MemoryDB`.

PR #8, opened 2019-01-28
update codec and new release

thiolliere:
This update codec and make new release of trie-bench and reference-tri

I only need trie-bench release for substrate so I don't know if we release a new one for reference-trie also.

as it is only dev depencencies of trie-root and trie-db we don't need to update them.


thiolliere:
(no other dependencies have major new version)

PR #10, opened 2019-02-18
Optim nibble encode

cheme:
This PR unroll the `at` in encode function of nibble_slice, it runs a bit faster but the code is less compact.


cheme:
closing it as #11 does not need that.

PR #11, opened 2019-02-19
Trie without extension node

cheme:
This PR implement new codec spec from polkadot-re-spec:

The main change is the removal of trie extension node. This is triggered by switching an associated constant in `TrieLayout`.
Other changes are:

* nibbleslice changes
 I did rework nibbleslice a bit: replacing `encode` by `left` and `right`.
 basically encode was copying and aligning bytes, this is mostly starting from
 the spec description.
 - left of a nibble offset is call `Prefix` and is encoded with padding at the end when 
 number of nibble does not fit
        this is used for prefix for key db : this is a breaking change but it got the nice property to allow iteration on key in db (for contract child trie for instance where we only got key of contract or in case kvdb evolves to allow collection specifics).
        Also note that this change in prefix also include the number of padding (needed to iterate correctly).
 - right of a nibble is call `Partial` this is similar to existing naming of variable and
   is what is encoded into trie nodes: the padding is at start.
   This allow leaf to avoid shifting bytes for its partial. That is not the case for branch.
   I believe leaf is more prone to use long partial (especially if key is resulting from a hash).
 - to avoid using a vec the types uses a slice and a optional last bytes (when padding).
   Actually the optional last bytes is (u8,u8) to allow other layout.

Note that some of the code could revert to slower usage of `at` function to be easier to code check.

 It also solve the existing TODO about removing header from encoding (see code in
 ethereum branch where a code impact is visible (in substrate it is all fine)).

* TrieLayout

An enveloppe trait over the trie layout to keep all implementation details at one place.
It also use experimental (few test) alternate trie layout (other radix only for aligned key and in a single byte
which limit to 4 variant).

* iter_build algorithm
  I write this at a time I did not see `trie_root` crate, I included it in the PR as I think it is not really hard to maintain and I used it a lot in testing.
  Note that this PR does not update `trie_root` to use TrieLayOut

* test branch from other project
https://github.com/cheme/parity-ethereum/tree/simple-codec : there are quite some change in codec, mainly related to the removal of nibble hpe
https://github.com/cheme/substrate/tree/simple-codec


cheme:
Note about `NodeHeader` scheme : compacting could look like:
```rust
// branch = value - 1 in extension size -> val 85 branch 84
const EMPTY_TRIE: u8 = 0;
const LEAF_NODE_OFFSET: u8 = 1; // (1 to 86) = max nbl l 85
const BRANCH_NODE_NODE_OFFSET: u8 = 87; // (87 to 171) max 84
const BRANCH_NODE_WITH_VALUE_OFFSET: u8 = 172; // (172 to 255) max 84
```
Problem, some scheme in substrate (double storage map) requires length > 32 bytes (from current code there is a 8 bit prefix + 128 bit twox hash + 256 bit black so a nibble length of 96 (more than the 85 allowed above)).
So the one byte length encoding does not fit. So it feels like it definitely should allow encoding on two bytes for big val.

hash-db/src/lib.rs
```diff
@@ -20,24 +20,32 @@
...
+/// An empty prefix constant. Mainly use for comparison
+/// and for root nodes.
+pub static EMPTY_PREFIX: Prefix<'static> = (&[], (0,0));
 
-/// Prefix byte information for avoding rc.
-pub type Prefix<'a> = &'a[u8];
+/// The Prefix of a trie node. Prefix is nibble path up to
+/// the node in the trie.
+/// For a value node it is the value key without the Partial
+/// bytes (encoded in the node).
```
arkpar: > For a value node it is the value key without the Partial bytes (encoded in the node).

So does this mean that the value node key has no prefix? Is it possible that two value nodes in different parts of the tree have the same partial key and therefore same hash?

hash-db/src/lib.rs
```diff
@@ -20,24 +20,32 @@
...
+/// Being the leftmost portion of a byte, its internal representation
```
arkpar: leftmost portion of a byte? You mean leftmost portion of the node key?

hash-db/src/lib.rs
```diff
@@ -20,24 +20,32 @@
...
+/// is byte slice and and a padded byte.
+/// The padded byte is represented as a pair with number of nibble first,
+/// and left aligned padded value second.
```
arkpar: This is a bit confusing. Why do we need to keep an additional nibble index here?

memory-db/src/lib.rs
```diff
@@ -160,19 +160,151 @@ impl<H, KF, T> Eq for MemoryDB<H, KF, T>
...
+		assert_eq!(
+			&ordered_prefixed_key::<KeccakHasher, Opt>(&k.clone().into(), (&p1[..], p2.clone())),
+			b,
+		);
+	}
+}
+/// Parameterize the size of successive chunks when using
+/// `ordered_prefixed_key`.
+/// Length of each chunk is minimum 1, maximum 254.
+pub trait OrderedPrefixedKeyChunk: IntoIterator<Item = usize> + Clone + MaybeDebug + Default {
```
arkpar: Whats a "chunk" and how is it formed? Furthermore, is this trait really needed? Does not seem to be implemented anywhere else.

memory-db/src/lib.rs
```diff
@@ -160,19 +160,151 @@ impl<H, KF, T> Eq for MemoryDB<H, KF, T>
...
-	let mut prefixed_key = Vec::with_capacity(key.as_ref().len() + prefix.len());
-	prefixed_key.extend_from_slice(prefix);
+///
+/// The scheme allows byte ordering of key depending on prefix (can iterate
+/// trie with it).
+/// Every chunk we add a 2 byte (could be reduce to one if storing directly a
+/// total number of chunks), the number of written bytes (last prefix padded one included), and
+/// the number of bytes in the prefix padding byte (note that for aligned prefix we add a
+/// theorically useless 0 to keep a simple trait).
+pub fn ordered_prefixed_key<H, C>(key: &H::Out, prefix: Prefix) -> Vec<u8>
```
arkpar: Do we really need this? Why is iteration not possible with current prefix scheme? It should be

test-support/reference-trie/src/lib.rs
```diff
@@ -17,21 +17,153 @@
...
 
-pub type RefTrieDB<'a> = trie_db::TrieDB<'a, keccak_hasher::KeccakHasher, ReferenceNodeCodec>;
-pub type RefTrieDBMut<'a> = trie_db::TrieDBMut<'a, KeccakHasher, ReferenceNodeCodec>;
-pub type RefFatDB<'a> = trie_db::FatDB<'a, KeccakHasher, ReferenceNodeCodec>;
-pub type RefFatDBMut<'a> = trie_db::FatDBMut<'a, KeccakHasher, ReferenceNodeCodec>;
-pub type RefSecTrieDB<'a> = trie_db::SecTrieDB<'a, KeccakHasher, ReferenceNodeCodec>;
-pub type RefSecTrieDBMut<'a> = trie_db::SecTrieDBMut<'a, KeccakHasher, ReferenceNodeCodec>;
-pub type RefLookup<'a, Q> = trie_db::Lookup<'a, KeccakHasher, ReferenceNodeCodec, Q>;
+/// trie layout similar to parity-ethereum
+pub struct LayoutOri;
```
arkpar: `LayoutOrigin`? Full name would be more readable

test-support/reference-trie/src/lib.rs
```diff
@@ -17,21 +17,153 @@
...
+	type H = keccak_hasher::KeccakHasher;
+	type C = ReferenceNodeCodec<BitMap16>;
+	type N = NibbleHalf;
+	type CB = Cache16;
+}
+
+impl TrieOps for LayoutOri { }
+
+/// trie layout similar to substrate one
+pub struct LayoutNew;
```
arkpar: "New" is used here and further in this PR is not descriptive. If we move to a different scheme this naming will get out of hand. Better come up with proper name now.

trie-db/src/lib.rs
```diff
@@ -351,29 +353,109 @@ where
...
 	}
 
 	/// Returns true iff the trie DB is a fat DB (allows enumeration of keys).
 	pub fn is_fat(&self) -> bool { self.spec == TrieSpec::Fat }
 }
+
+/// Trait with definition of trie layout.
+/// It contains all associated trait needed for
+/// a trie definition.
+pub trait TrieLayOut {
```
arkpar: `TrieLayout`

test-support/reference-trie/src/lib.rs
```diff
@@ -17,21 +17,153 @@
...
+impl TrieLayOut for LayoutNew {
+	const USE_EXTENSION: bool = false;
+	type H = keccak_hasher::KeccakHasher;
+	type C = ReferenceNodeCodecNoExt<BitMap16>;
+	type N = NibbleHalf;
+	type CB = Cache16;
+}
+
+/// trie layout similar to substrate one
+pub struct LayoutNewH<H>(PhantomData<H>);
```
arkpar: `SimpleHashingLayout` perhaps?

test-support/reference-trie/src/lib.rs
```diff
@@ -17,21 +17,153 @@
...
+	}
+}
+
+/// bitmap codec for radix 4
+pub struct BitMap4(u8);
+
+impl ChildBitmap for BitMap4 {
+	const ENCODED_LEN: usize = 1;
+	type Error = ReferenceError;
+	type Buff = [u8;2]; // need a byte for header
```
arkpar: `Buffer`

test-support/reference-trie/src/lib.rs
cheme: Yes, I am also removing the idea of new and origin, I will use 'ExtensionLayout' and 'NoExtensionLayout'. The fact that reference trie uses same scheme as substrate should not be something strict (code will diverge at some point).

memory-db/src/lib.rs
cheme: We do not really need this (this implementation is linked nowhere and I can consider removing it).

Current prefix does not allow direct kvdb iteration because we cannot say if we are comparing prefix or node hash, so we need to encode length of the prefix somehow, but encoding it tends to break the simple byte array comparison thus this complicated scheme.
But the main point is that being able to iterate on the trie nodes from kvdb have many limitation that makes it not interesting in most case:
- for state it iterate over all state nodes (we can see the history of value with a same prefix but helas it is not the same as value with same key), so for a collection containing multiple state of a trie it is not really usefull.
- for trie that got a single state in a single collection (or a with a common prefix), the iteration should work.

hash-db/src/lib.rs
cheme: I realize the additional nibble index does not need to be use in the prefix, but it seems like it can be good to have all info, especially since it is use for a trait (for comparing prefix it is needed but that is not really needed).

It also allows building nibble from prefix (strictly speaking it should not be necessary as prefix is either slice and padded byte from key or from a nibblevec).

hash-db/src/lib.rs
cheme: Same partial key and same hash is a possibility yes, but at that point either the prefix are different or the key are the same (key = prefix ++ partial).
Does it mean that the value node key has no prefix? Ah ok, I will rephrase that.

test-support/reference-trie/src/lib.rs
rphmeier: > similar to substrate one

that's not a very helpful comment from the perspective of this codebase

memory-db/src/lib.rs
arkpar: I'd remove it. There's no use case for querying node by hash or iterating all nodes I believe. All we need is tree iteration, that is walking the tree starting from a root node down, not just enumerating all nodes. And that should work with unordered prefix.

test-support/reference-trie/src/lib.rs
```diff
@@ -51,6 +207,32 @@ const EXTENSION_NODE_OVER: u8 = BRANCH_NODE_NO_VALUE - EXTENSION_NODE_OFFSET;
...
+	pub const LEAF_NODE_OVER: u8 = LEAF_NODE_BIG - LEAF_NODE_OFFSET;
+	pub const BRANCH_NODE_WITH_VALUE_OVER: u8 = BRANCH_NODE_WITH_VALUE_BIG - BRANCH_NODE_WITH_VALUE;
+	pub const BRANCH_NODE_NO_VALUE_OVER: u8 = BRANCH_NODE_NO_VALUE_BIG - BRANCH_NODE_NO_VALUE;
+	pub const LEAF_NODE_LAST: u8 = LEAF_NODE_BIG - 1;
+	pub const BRANCH_NODE_WITH_VALUE_LAST: u8 = BRANCH_NODE_WITH_VALUE_BIG - 1;
+	pub const BRANCH_NODE_NO_VALUE_LAST: u8 = BRANCH_NODE_NO_VALUE_BIG - 1;
+}*/
+
+/// constant use with trie simplification codec
+mod s_cst {
```
arkpar: This does not help with readability. Modules should be used to group constants (e.g. `mask` for the last three) or not used at all.

test-support/reference-trie/src/lib.rs
```diff
@@ -147,21 +434,127 @@ impl Encode for NodeHeader {
 	}
 }
 
+fn s_size_and_prefix_iter(size: usize, prefix: u8) -> impl Iterator<Item = u8> {
```
arkpar: What's is the `s_` prefix here and in other function names below?

test-support/reference-trie/src/lib.rs
```diff
@@ -67,19 +249,52 @@ fn fuse_nibbles_node<'a>(nibbles: &'a [u8], leaf: bool) -> impl Iterator<Item =
...
+
+	let iter_start = match kind {
+		NodeKindNoExt::Leaf => s_size_and_prefix_iter(size, s_cst::LEAF_PREFIX_MASK),
+		NodeKindNoExt::BranchNoValue => s_size_and_prefix_iter(size, s_cst::BRANCH_WITHOUT_MASK),
+		NodeKindNoExt::BranchWithValue => s_size_and_prefix_iter(size, s_cst::BRANCH_WITH_MASK),
+	};
+	iter_start
+		.chain(if nibbles.len() % 2 == 1 { Some(nibbles[0]) } else { None })
+		.chain(nibbles[nibbles.len() % 2..].chunks(2).map(|ch| ch[0] << 4 | ch[1]))
+		//.chain(nibbles[..nibbles.len() - (nibbles.len() % 2)].chunks(2).map(|ch| ch[0] << 4 | ch[1]))
```
arkpar: Our policy is not to commit commented out code

test-support/reference-trie/src/lib.rs
```diff
@@ -51,6 +207,32 @@ const EXTENSION_NODE_OVER: u8 = BRANCH_NODE_NO_VALUE - EXTENSION_NODE_OFFSET;
...
+	pub const BRANCH_NODE_NO_VALUE_OVER: u8 = BRANCH_NODE_NO_VALUE_BIG - BRANCH_NODE_NO_VALUE;
+	pub const LEAF_NODE_LAST: u8 = LEAF_NODE_BIG - 1;
+	pub const BRANCH_NODE_WITH_VALUE_LAST: u8 = BRANCH_NODE_WITH_VALUE_BIG - 1;
+	pub const BRANCH_NODE_NO_VALUE_LAST: u8 = BRANCH_NODE_NO_VALUE_BIG - 1;
+}*/
+
+/// constant use with trie simplification codec
+mod s_cst {
+	pub const EMPTY_TRIE: u8 = 0;
+	pub const NIBBLE_SIZE_BOUND: usize = u16::max_value() as usize;
```
arkpar: `MAX_KEY_NIBBLES` is more appropriate. NIbble size is always 4 bits.

test-support/reference-trie/src/lib.rs
```diff
@@ -191,69 +584,126 @@ fn take<'a>(input: &mut &'a[u8], count: usize) -> Option<&'a[u8]> {
...
-		output.push(partial[0] & 0x0f);
+	if number_nibble_encoded > 0 {
+		output.push(N::masked_right(number_nibble_encoded as u8, (partial.0).1));
+	}
+	output.extend_from_slice(&partial.1[..]);
+	output
+}
+
+
+fn partial_to_key_it<N: NibbleOps, I: Iterator<Item = u8>>(partial: I, nibble_count: usize, offset: u8, over: u8) -> Vec<u8> {
```
arkpar: `partial_iter_to_key`

test-support/reference-trie/src/lib.rs
```diff
@@ -191,69 +584,126 @@ fn take<'a>(input: &mut &'a[u8], count: usize) -> Option<&'a[u8]> {
...
+	assert!(nibble_count < over as usize);
+	let mut output = Vec::with_capacity(1 + (nibble_count / N::NIBBLE_PER_BYTE));
+	output.push(offset + nibble_count as u8);
+	output.extend(partial);
+	output
+}
+
+fn partial_enc_it<N: NibbleOps, I: Iterator<Item = u8>>(partial: I, nibble_count: usize, node_kind: NodeKindNoExt) -> Vec<u8> {
```
arkpar: `partial_iter_encode` ?

gavofyork:
@cheme please give this a thorough review; there are several points on code-quality that were made in other reviews, particularly cryptic naming and low-quality or non-existent documentation that are in here.

test-support/reference-trie/src/lib.rs
cheme: codec can be use with other length of nibble size (TtrieStream cannot), I am starting to consider removing this possibility through (I did not run enough test on other trie radix).

gavofyork:
@cheme when you're certain that *both* all the reported issues, *and the typical issues* (bad naming, missing docs, low-quality docs) are fixed, please ping @arkpar and @svyatonik to get a final review.

memory-db/src/lib.rs
```diff
@@ -160,19 +160,151 @@ impl<H, KF, T> Eq for MemoryDB<H, KF, T>
...
+}
+
+
+/// Make database key from hash and prefix.
+///
+/// Prefix could be ordered, but key hash appending will break this property.
+/// If true ordering is needed we would need a fix length encoding for prefix
+/// (so a maximum length and additional byte (or chunk of prefix encoded)).
+pub fn prefixed_key<H: KeyHasher>(key: &H::Out, prefix: Prefix) -> Vec<u8> {
+	let mut prefixed_key = Vec::with_capacity(key.as_ref().len() + prefix.0.len() + 1);
```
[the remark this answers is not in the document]
cheme: I removed the nibble length : my reasoning for its need was wrong: two nodes using this prefix (one with padding and one without are in the same trie path so if both got the same hash it would indicate a trie loop or a hash collision).

cheme:
@arkpar, @svyatonik I went other bad naming and try to go over the doc again this morning.

There is still things I can do to reduce code review size:

- remove fuzzing and bench.
- remove iter_build (extensively use by fuzzing but not much at other place: I may even put it in its own crate (there may be some codec thing that will need to go public)).
- remove other radix variant, this can simplify nibble logic a bit (main idea of this was to make things more clear by generalizing but when I read back the code I am not sure its).
- remove one of the two encoding variant from test-support.

gavofyork:
@cheme please remove as much as you can to keep the review size down

gavofyork:
@arkpar @svyatonik would be great if we could get this reviewed again

arkpar:
Fuzzing and benching may be left as is.
Generic nibble size/tree arity sounds nice, but in practice also introduces additional overhead. E.g. the need to store extra byte per key prefix. I'd remove it for now.

test-support/reference-trie/src/lib.rs
```diff
@@ -15,32 +15,129 @@
...
+impl<H: Hasher> TrieLayout for GenericNoExtensionLayout<H> {
+	const USE_EXTENSION: bool = false;
+	type Hash = H;
+	type Codec = ReferenceNodeCodecNoExt;
+}
+
+impl<H: Hasher> TrieConfiguration for GenericNoExtensionLayout<H> { }
+
+/// Trie layout without extension nodes.
+pub struct NoExtensionLayout;
```
svyatonik: Could probably be `pub type NoExtensionLayout = GenericNoExtensionLayout<keccak_hasher::KeccakHasher>;`?

trie-db/src/iter_build.rs
```diff
@@ -0,0 +1,803 @@
...
+pub trait CacheBuilder<HO> {
+	/// Size of cache.
+	const SIZE: usize;
+	/// The type of the cache.
+	type Cache: AsRef<[CacheNode<HO>]> + AsMut<[CacheNode<HO>]>;
+	/// Create a new cache.
+	fn new_vec_slice_buffer() -> Self::Cache;
+}
+
+/// Cache builder for radix 16 trie.
```
svyatonik: Are `Cache4` and `Cache16` still used anywhere?

gavofyork:
@cheme couple of questions there...

@arkpar @svyatonik are you guys ok with the PR otherwise?

trie-db/src/nibble/mod.rs
```diff
@@ -0,0 +1,192 @@
...
+	/// The nibble shifts needed to align.
+	/// We use two value, one is a left shift and
+	/// the other is a right shift.
+	pub const SPLIT_SHIFTS: (usize, usize) = (4, 4);
+
+	/// Count the biggest common depth between two left aligned packed nibble slice.
+	pub fn biggest_depth(v1: &[u8], v2: &[u8]) -> usize {
+		// sorted assertion preventing out of bound
+		for a in 0..v1.len() {
+			if v1[a] == v2[a] {
```
svyatonik: Is that always correct that `v1.len() <= v2.len()` (otherwise it'll panic)? I have found the only usage of the `biggest_depth()` from the `trie_visit()`. I assume that `input` there is ordered in some manner? If it sorted using `BTreeMap`, then given that e.g. `vec![5, ,5] < vec![20]`, this could panic, no?

arkpar:
I've verified that the implementation matches the spec and there's sufficient test coverage. So "looksgood" from me.

trie-db/src/nibble/mod.rs
cheme: In fact there is no verification of the ordering, iter_build code does not support unordered input.
I can prevent this failing, but I will need to propagate the error, that seems pretty quick to do.

svyatonik: Sorry - I mean that even if input is ordered (lexicographically, I suppose), then this could panic, because lexicographically vec of larger len could came before vec of smaller len. No need to propagate error imo.

cheme: yes (I just realized, I just got to min the loop bound).

svyatonik: (I think you also need to update this line: `return v1.len() * NIBBLE_PER_BYTE;` below to use `cmp::min()` instead of `v1.len()`)

PR #12, opened 2019-02-20
Node key prefixes in the database

arkpar:
For some applications, such as substrate it is desirable to have each node to be unique in the trie. So that the same node can't be inserted into two separate branches of the same trie. This simplifies implementation of node storage quite a lot, since it removes the need for reference counting.

In ethereum the uniqueness is already guaranteed by the fact that the keys are hashes and each value ends up in a leaf node with a long random partial key. In Substrate this is not the case, as the keys are plain.

This PR introduces an additional parameter for `HashDB` functions that takes encoded partial node key. This allows for separating colliding nodes on the database levels, and for more efficient database implementation. E.g. The nodes that are close to each other in the trie may be grouped on disk.
Trie root calculation is not affected.

### DIFF
--- a/memory-db/src/lib.rs
+++ b/memory-db/src/lib.rs
@@ -24,8 +24,6 @@ use std::collections::hash_map::Entry;
 use std::collections::HashMap;
 use std::mem;
 
-type Key = Vec<u8>;
-
 /// Reference-counted memory-based `HashDB` implementation.
 ///
 /// Use `new()` to create a new database. Insert items with `insert()`, remove items
@@ -41,9 +39,9 @@ type Key = Vec<u8>;
 ///
 /// use hash_db::{Hasher, HashDB};
 /// use keccak_hasher::KeccakHasher;
-/// use memory_db::MemoryDB;
+/// use memory_db::{MemoryDB, HashKey};
 /// fn main() {
-///   let mut m = MemoryDB::<KeccakHasher, Vec<u8>>::default();
+///   let mut m = MemoryDB::<KeccakHasher, HashKey<_>, Vec<u8>>::default();
 ///   let d = "Hello world!".as_bytes();
 ///
 ///   let k = m.insert(&[], d);
@@ -74,26 +72,76 @@ type Key = Vec<u8>;
 /// }
 /// ```
 #[derive(Clone, PartialEq)]
-pub struct MemoryDB<H: KeyHasher, T> {
-	data: HashMap<Key, (T, i32)>,
+pub struct MemoryDB<H, KF, T>
+	where
+	H: KeyHasher,
+	KF: KeyFunction<H>,
+{
+	data: HashMap<KF::Key, (T, i32)>,
 	hashed_null_node: H::Out,
 	null_node_data: T,
+	_kf: ::std::marker::PhantomData<KF>,
 }
 
-impl<'a, H, T> Default for MemoryDB<H, T>
+pub trait KeyFunction<H: KeyHasher> {
+	type Key: Send + Sync + Clone + std::hash::Hash + std::cmp::Eq ;
+
+	fn key(hash: &H::Out, prefix: &[u8]) -> Self::Key;
+}
+
+
+/// Make database key from hash and prefix.
+pub fn prefixed_key<H: KeyHasher>(key: &H::Out, prefix: &[u8]) -> Vec<u8> {
+	let mut prefixed_key = Vec::with_capacity(key.as_ref().len() + prefix.len());
+	prefixed_key.extend_from_slice(prefix);
+	prefixed_key.extend_from_slice(key.as_ref());
+	prefixed_key
+}
+
+/// Make database key from hash only.
+pub fn hash_key<H: KeyHasher>(key: &H::Out, _prefix: &[u8]) -> H::Out {
+	key.clone()
+}
+
+/// Key function that only uses the hash
+pub struct HashKey<H: KeyHasher>(std::marker::PhantomData<H>);
+
+impl<H: KeyHasher> KeyFunction<H> for HashKey<H> {
+	type Key = H::Out;
+
+	fn key(hash: &H::Out, prefix: &[u8]) -> H::Out {
+		hash_key::<H>(hash, prefix)
+	}
+}
+
+/// Key function that concatenates prefix and hash.
+pub struct PrefixedKey<H: KeyHasher>(std::marker::PhantomData<H>);
+
+impl<H: KeyHasher> KeyFunction<H> for PrefixedKey<H> {
+	type Key = Vec<u8>;
+
+	fn key(hash: &H::Out, prefix: &[u8]) -> Vec<u8> {
+		prefixed_key::<H>(hash, prefix)
+	}
+}
+
+impl<'a, H, KF, T> Default for MemoryDB<H, KF, T>
 where
 	H: KeyHasher,
-	T: From<&'a [u8]>
+	T: From<&'a [u8]>,
+	KF: KeyFunction<H>,
 {
 	fn default() -> Self {
 		Self::from_null_node(&[0u8][..], [0u8][..].into())
 	}
 }
 
-impl<H, T> MemoryDB<H, T>
+/// Create a new `MemoryDB` from a given null key/data
+impl<H, KF, T> MemoryDB<H, KF, T>
 where
 	H: KeyHasher,
 	T: Default,
+	KF: KeyFunction<H>,
 {
 	/// Remove an element and delete it from storage if reference count reaches zero.
 	/// If the value was purged, return the old value.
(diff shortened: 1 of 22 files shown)


trie-db/src/triedbmut.rs
```diff
@@ -826,7 +827,7 @@ where
 
 		match self.storage.destroy(handle) {
 			Stored::New(node) => {
-				let encoded_root = node.into_encoded::<_, C, H>(|child| self.commit_child(child) );
+				let encoded_root = node.into_encoded::<_, C, H>(|child, k| self.commit_child(child, k));
```
cheme: Would it make sense to also apply the scramble to root (with empty partial key)? (a few line below)

cheme:
Looks interesting, out of curiosity, why is this needed ? (`TrieStream` of trie-root probably also need an updated).

trie-db/src/triedbmut.rs
```diff
@@ -131,20 +131,21 @@ where
...
 			Node::Branch(mut children, value) => {
 				C::branch_node(
 					// map the `NodeHandle`s from the Branch to `ChildReferences`
 					children.iter_mut()
 						.map(Option::take)
-						.map(|maybe_child|
-							maybe_child.map(|child| child_cb(child))
+						.enumerate()
+						.map(|(i, maybe_child)|
+							maybe_child.map(|child| child_cb(child, &NibbleSlice::new(&[i as u8]).encoded(false)))
```
cheme: From what I understand of the issue, I would scramble with full key instead of partial key (just moving cursor over the key nibble slice while going down the tree). I am afraid of the case where node are the same with same partial key to.

arkpar:
Updated the description.

trie-db/src/triedbmut.rs
arkpar: That's already the case. `child_cb` accepts node's partial key and combines it with parent's full key and passes it down the line.

cheme: :+1: I missed that, smart (the cursor thing does not work for full trie recursive calculation I forgot it is done on commit).

xlc:
@cheme Anything prevents this from been merged? I really want https://github.com/paritytech/substrate/issues/1733 get fixed.

cheme:
@xlc, the code in itself seems functionally fine and could be merge, but it is currently unclear if this will be the actual fix for https://github.com/paritytech/substrate/issues/1733, it would also require trie_root changes for full compatibility (probably requiring modifying 'TrieStream' trait).
Using this alternate scheme changes all tries content (every hash ref in every node so also every root) : as I see it, for substrate it means either implementing a way to run two version of trie depending on block (quite difficult), or rebooting the chain.
I believe @arkpar is working on an alternate fix to avoid such breaking change, I do not have the details (for my part I can only imagine using key as immediate backing db key for storage, which would require a db migration (could improve perf a bit for fewer operation on update of value but I did not think to much about this)).
So I totally agree with you that 1733 is a very high priority issue, but using time on alternate possible solution could really be worth it.

xlc:
@cheme Thanks for the detailed explanation. Good to know this is not getting stale.

memory-db/src/lib.rs
```diff
@@ -284,52 +282,89 @@ where
 	fn contains(&self, key: &H::Out) -> bool { PlainDB::contains(self, key) }
 }
 
+
+/// Make database key from hash and prefix.
+pub fn full_key<H: KeyHasher>(key: &H::Out, prefix: &[u8]) -> Key {
+	let mut full_key = Vec::with_capacity(key.as_ref().len() + prefix.len());
+	full_key.extend_from_slice(prefix);
+	full_key.extend_from_slice(key.as_ref());
```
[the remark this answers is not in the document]
arkpar: Profiling shows that allocations here and calculating partial keys are insignificant compared to IO and other code. So I left it for later.

memory-db/src/lib.rs
```diff
@@ -100,11 +97,12 @@ where
 {
 	/// Remove an element and delete it from storage if reference count reaches zero.
 	/// If the value was purged, return the old value.
-	pub fn remove_and_purge(&mut self, key: &<H as KeyHasher>::Out) -> Option<T> {
+	pub fn remove_and_purge(&mut self, key: &<H as KeyHasher>::Out, prefix: &[u8]) -> Option<T> {
```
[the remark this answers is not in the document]
arkpar: That's a good point. I'll see if I can make the key function parameter to `MemoryDB`

memory-db/src/lib.rs
```diff
@@ -284,52 +282,89 @@ where
 	fn contains(&self, key: &H::Out) -> bool { PlainDB::contains(self, key) }
 }
 
+
+/// Make database key from hash and prefix.
+pub fn full_key<H: KeyHasher>(key: &H::Out, prefix: &[u8]) -> Key {
```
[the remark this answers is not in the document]
arkpar: Reusing the buffer would only make sense in read-only methods, and that would make `MemoryDB` not thread safe for reading, or require additional synchronization.

trie-db/src/triedbmut.rs
```diff
@@ -430,21 +431,21 @@ where
 	}
 
 	/// insert a key-value pair into the trie, creating new nodes if necessary.
-	fn insert_at(&mut self, handle: NodeHandle<H::Out>, partial: NibbleSlice, value: DBValue, old_val: &mut Option<DBValue>) -> Result<(StorageHandle, bool), H::Out, C::Error> {
+	fn insert_at(&mut self, handle: NodeHandle<H::Out>, partial: NibbleSlice, key: NodeKey, value: DBValue, old_val: &mut Option<DBValue>) -> Result<(StorageHandle, bool), H::Out, C::Error> {
```
cheme: In trie_db_mut.rs ,  function arguments  `partial: NibbleSlice, key:NodeKey` could be fused in a single struct (operation on both seems symetric: key increase from what partial decrease on recursive calls at some/many place).
eg
```rust
let child_key = combine_encoded(&key, &partial.encoded_leftmost(cp, false));
match self.remove_at(child_branch, partial.mid(cp), child_key.clone(), old_val)? {
```
do not know if it is worth a new struct, just mentioning it.

cheme:
I did start updating another trie pr on friday using this pr changes: things seems to work fine :)  (after solving few indexing issue of my own, I could fuzz my triebuilder against this prefixed implementation). Manipulating this new api makes me wonder about two points:

- for https://github.com/paritytech/substrate/issues/2035 , I started overloading db implementation (for substrate usage of triedb and triedbmut in child trie context) by prefixing key with a keyspace. If `KeyFunction` trait changes to :
```rust
fn key(&self, hash: &H::Out, prefix: &[u8]) -> Self::Key; 
```
I could make keyspace indexing in the keyfunction implementation.
This implies that triedb and triedbmut will build with a third parameter (keyfunction struct as inner field), but it won't really add verbosity (we already need to set keyfunction type when building trie and triemut).
This could also be use to add additional info in the key for some particular trie usage (like block_number|prefix|hash or fix_size_prefix|block_number|hash  (triedb and triedbmut keyfunction struct needs obviously in this case to contain a reference to the block_number)). 

But honestly, this is still doable by overloading the db if we consider this not being the role of the keyfunction trait.

- I am not really comfortable with using nibble trie encoding for db key prefix (it does not allow db sorted iteration over a common prefixed key in the db due to the header infos).
Ideally the way we encode prefix in db could be defined into memorydb: into `KeyFunction` implementation. It would requires to pass Key plus nibble index (or an equivalent to nibbleslice) to the key function instead of the encoded prefix.
It can also avoid running the nibble encoding in the parity eth case (shouldn't be the reason to do it this way).

arkpar:
Regarding paritytech/substrate#2035, this could be done purely in the `HashDb` implementation. `KeyFunction` is implementation detail of `MemoryDb`. I don't think it should be part of the `TrieDb` or `TrieDbMut`. The trie itself is not concerned with duplicate nodes or reference counting or storage optimizations. It just provide enough information for the backend to resolve all these issues.

> I am not really comfortable with using nibble trie encoding for db key prefix (it does not allow db sorted iteration over a common prefixed key in the db due to the header infos).

Not sure I understand this. Trie iteration is surely still possible. Node backend iteration or seek depends on `HashDb` implementation and out of the scope of this PR. E.g. with substrate key function iteration is still possible, but seeking by node hash is not.

cheme:
About paritytech/substrate#2035 , the HashDB way of doing things is fine (still I found it more elegant with KeyFunction (I am currently realizing that I need to handle the empty node value of hash db case : with keyfunction it is native). But it is really a matter of design and does not have to make it to this PR.
About iteration, it is mostly speculating about db (rocksdb or other with ordered tree iteration) optimization : having a common prefix looks better for me but as for the previous point it does not have to make it to this pr.

PR #13, opened 2019-02-27
No std support for trie

cheme:
This PR allows trie to run in no_std mode : eg https://github.com/paritytech/substrate/compare/master...cheme:no_std-trie 
It requires 'alloc' so it is no_std for nightly only.
This replies to https://github.com/paritytech/substrate/issues/1574 .

For memory-db and trie I use 'hashmap_core' crate, this is not a very good choice as 'hashbrown' is more maintained, and long term it should be replaced by 'hashbrown'. I did not use 'hashbrown' because the way we switch to 'no_std' is done through feature declaration (feature 'nightly'), that means that I have to propagate the feature up to the no_std building crate (see commit history of this pr) which is very awkward.

Versioning update is minor and global (I believe it is a choice to keep versioning constistent between trie subproject).

I kept some Debug implementation (basically to use 'expect' on result), I do not think it is too heavy for no_std but it could be a bad choice and we could also avoid 'expect' call in code. Some Debug code is not include but could probably be (full trie debug).

### DIFF
--- a/hash-db/Cargo.toml
+++ b/hash-db/Cargo.toml
@@ -1,11 +1,15 @@
 [package]
 name = "hash-db"
-version = "0.9.0"
+version = "0.9.1"
 authors = ["Parity Technologies <admin@parity.io>"]
 description = "Trait for hash-keyed databases."
 license = "Apache-2"
 categories = [ "no-std" ]
 
+[dependencies]
+hashbrown = { version = "0.1" }
+
 [features]
 default = ["std"]
-std = []
+std = [
+]
--- a/hash-db/Cargo.toml
+++ b/hash-db/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "hash-db"
-version = "0.11.0"
+version = "0.11.1"
 authors = ["Parity Technologies <admin@parity.io>"]
 description = "Trait for hash-keyed databases."
 license = "Apache-2.0"
--- a/trie-db/src/lib.rs
+++ b/trie-db/src/lib.rs
@@ -41,6 +41,7 @@ extern crate memory_db;
 extern crate keccak_hasher;
 #[cfg(all(feature = "std", test))]
 extern crate reference_trie;
+
 #[cfg(not(feature = "std"))]
 extern crate hashbrown;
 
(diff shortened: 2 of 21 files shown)


hash-db/Cargo.toml
```diff
@@ -1,11 +1,12 @@
...
+version = "0.12.2"
 authors = ["Parity Technologies <admin@parity.io>"]
 description = "Trait for hash-keyed databases."
 license = "Apache-2.0"
 categories = [ "no-std" ]
 
 [features]
 default = ["std"]
-std = []
+std = [
```
bkchr: Why this split between two lines?

trie-db/src/node_codec.rs
```diff
@@ -19,12 +19,25 @@ use hash_db::Hasher;
...
+use std::error::Error;
+
+
+#[cfg(not(feature = "std"))]
+use alloc::vec::Vec;
+
+use elastic_array::ElasticArray128;
+
+#[cfg(not(feature = "std"))]
+pub trait Error: core::fmt::Debug {}
```
bkchr: Requiring `Debug` on `no_std` does not sound that good.

trie-db/src/lib.rs
```diff
@@ -32,11 +39,41 @@ extern crate trie_root;
...
+
+use core_::{fmt, marker::PhantomData};
+
+#[cfg(feature = "std")]
+use std::error::Error;
+
+#[cfg(feature = "std")]
+use std::fmt::Debug;
+#[cfg(feature = "std")]
+pub trait DebugIfStd: Debug {}
```
bkchr: Normally this is called `MaybeDebug`.

trie-db/src/lib.rs
```diff
@@ -80,7 +117,7 @@ pub enum TrieError<T, E> {
 	DecoderError(T, E),
 }
 
-impl<T, E> fmt::Display for TrieError<T, E> where T: std::fmt::Debug, E: std::fmt::Debug {
+impl<T, E> fmt::Display for TrieError<T, E> where T: ::core_::fmt::Debug, E: ::core_::fmt::Debug {
```
bkchr: Why do you here `Debug` directly? No `DebugIfStd/MaybeDebug`?

trie-db/src/node_codec.rs
cheme: Yes it does not look super good, I did test, I enabled it to be able to use 'expect' in code (it is only 4 expect but if we do not implement it it seems super likely that we break no_std in next trie pr). I still can remove those 4 expect, what do you think?

bkchr: It is just that `Debug` can generate a lot of code that blows up the WASM size.

PR #15, opened 2019-03-28
Fix iteration for prefixed storage

arkpar:
Trie iteration was not implemented correctly in #12 
Also switched tests to use prefixed storage.

### DIFF
--- a/trie-db/src/triedb.rs
+++ b/trie-db/src/triedb.rs
@@ -90,10 +90,11 @@ where
 			.ok_or_else(|| Box::new(TrieError::InvalidStateRoot(*self.root)))
 	}
 
-	/// Given some node-describing data `node`, return the actual node RLP.
+	/// Given some node-describing data `node`, and node key return the actual node RLP.
 	/// This could be a simple identity operation in the case that the node is sufficiently small, but
 	/// may require a database lookup. If `is_root_data` then this is root-data and
 	/// is known to be literal.
+	/// `partial_key` is encoded nibble slice that addresses the node.
 	fn get_raw_or_lookup(&'db self, node: &[u8], partial_key: &[u8]) -> Result<Cow<'db, DBValue>, H::Out, C::Error> {
 		match (partial_key == nibbleslice::EMPTY_ENCODED, C::try_decode_hash(node)) {
 			(false, Some(key)) => {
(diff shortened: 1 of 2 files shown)


trie-db/src/triedb.rs
```diff
@@ -95,7 +95,7 @@ where
 	/// may require a database lookup. If `is_root_data` then this is root-data and
 	/// is known to be literal.
 	fn get_raw_or_lookup(&'db self, node: &[u8], partial_key: &[u8]) -> Result<Cow<'db, DBValue>, H::Out, C::Error> {
```
rphmeier: i think it could be helpful to have a doc here about `partial_key` and how it's expected to be an encoded nibble-slice.

trie-db/src/triedb.rs
```diff
@@ -95,7 +95,7 @@ where
...
-		match (partial_key.is_empty(), C::try_decode_hash(node)) {
+		match (partial_key == nibbleslice::EMPTY_ENCODED, C::try_decode_hash(node)) {
```
rphmeier: a bit beyond the scope of the PR, but is this `match` assuming that only empty partial-keys could lead to nodes small enough to fit in the hash? A small partial key would easily break that if so

trie-db/src/triedb.rs
```diff
@@ -371,6 +373,17 @@ impl<'a, H: Hasher, C: NodeCodec<H>> TrieDBIterator<'a, H, C> {
 		}
 		result
 	}
+
+	/// Encoded key for storage lookup
+	fn encoded_key(&self) -> ElasticArray36<u8> {
```
rphmeier: is this not also encoded somewhere else?

trie-db/src/triedb.rs
arkpar: Previously (before #12) there was a check for root node here.

trie-db/src/triedb.rs
arkpar: This is a method of `TrieDBIterator`, which manages current key as a vector of nibbles. I considered switching to `NibbleVec` instead, that uses compact form and can be trivially converted to `NibbleSlice`. Unfortunately the conversion would only work for even number of nibbles, and it is not trivial to fix that.

trie-db/src/triedb.rs
rphmeier: i see, maybe a comment about that?

PR #20, opened 2019-06-18
fix no_std build for memory-db.

cheme:
Version 12.3 broke no_std compatibility.
This PR fixes that.
The PR also removes some warning (add 'dyn' info on types).

### DIFF
--- a/memory-db/src/lib.rs
+++ b/memory-db/src/lib.rs
@@ -389,12 +389,9 @@ where
 		use core::mem::size_of;
 		let mut n = self.data.capacity() * (size_of::<T>() + size_of::<H>() + size_of::<usize>());
 		for (k, v) in self.data.iter() {
-			n += k.size_of(ops);
-			n += v.size_of(ops);
+			n += k.size_of(ops) + v.size_of(ops);
 		}
-		n
-			+ self.null_node_data.size_of(ops)
-			+ self.hashed_null_node.size_of(ops)
+		n + self.null_node_data.size_of(ops) + self.hashed_null_node.size_of(ops)
 	}
 }
 
--- a/memory-db/src/lib.rs
+++ b/memory-db/src/lib.rs
@@ -354,6 +354,9 @@ where
 	}
 }
 
+// not no_std implementation requires that hasmap core
+// got its implementation in parity-util-mem
+#[cfg(feature = "std")]
 impl<H, KF, T> MallocSizeOf for MemoryDB<H, KF, T>
 where
 	H: KeyHasher,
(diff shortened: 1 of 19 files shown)


memory-db/src/lib.rs
```diff
@@ -369,6 +372,33 @@ where
...
+	KF::Key: MallocSizeOf,
+{
+	fn size_of(&self, ops: &mut MallocSizeOfOps) -> usize {
+		use core::mem::size_of;
+		let mut n = self.data.capacity() * (size_of::<T>() + size_of::<H>() + size_of::<usize>());
+		for (k, v) in self.data.iter() {
+			n += k.size_of(ops);
+			n += v.size_of(ops);
+		}
+		n
```
bkchr: Why these line breaks? (I think all should fit on one line)

memory-db/src/lib.rs
```diff
@@ -369,6 +372,33 @@ where
...
+	H::Out: MallocSizeOf,
+	T: MallocSizeOf,
+	KF: KeyFunction<H>,
+	KF::Key: MallocSizeOf,
+{
+	fn size_of(&self, ops: &mut MallocSizeOfOps) -> usize {
+		use core::mem::size_of;
+		let mut n = self.data.capacity() * (size_of::<T>() + size_of::<H>() + size_of::<usize>());
+		for (k, v) in self.data.iter() {
+			n += k.size_of(ops);
```
bkchr:
```suggestion
			n += k.size_of(ops) + v.size_of(ops);
```

memory-db/src/lib.rs
```diff
@@ -369,6 +372,33 @@ where
...
+			n += v.size_of(ops);
```
bkchr:
```suggestion
```

memory-db/src/lib.rs
```diff
@@ -354,6 +354,9 @@ where
 	}
 }
 
+// not no_std implementation requires that hasmap core
+// got its implementation in parity-util-mem
```
dvdplm: This comment is unclear to me, can you rephrase it? Is it "The `no_std` implementation requires that `Hashmap` has an impl of `MallocSizeOf` in `parity-util-mem`" maybe?

memory-db/src/lib.rs
```diff
@@ -369,6 +372,30 @@ where
 	}
 }
 
+// This is temporary code and should be in
+// parity-util-mem, but hashmap_core will probably
+// get replaced at some point by hashbrown so
+// this is a pragmatic temporary solution.
```
dvdplm: Can you open an issue so we don't lose track of this?

dvdplm:
@cheme Will you tag&publish also?

cheme:
yes and yank 12.3 (I also will update the parity pr with heapsize removal).

PR #25, opened 2019-07-29
Update codec + new release of trie-bench

thiolliere:
We just need new version of trie-bench.

and reference-trie desn't need a major version bump because last version is not published


thiolliere:
hmm maybe at some point we could improve error in reference-trie as in https://github.com/paritytech/substrate/pull/3232/files#diff-b11b301b92d34723219911fa5b34ace9

PR #31, opened 2019-10-17
export Error for no_std

yjhmelody:
I compile `substrate/core/trie` with no_std.
I found the Error in `no_std` is not exported by trie-db.
And `core/trie` seems can not be compiled to `no_std` now time.
```
 core/trie/src/node_codec.rs:43:17
   |
43 | impl<H: Hasher> NodeCodecT<H> for NodeCodec<H> {
   |                 ^^^^^^^^^^^^^ the trait `std::error::Error` is not implemented for `error::Error`

```

### DIFF
--- a/trie-db/src/lib.rs
+++ b/trie-db/src/lib.rs
@@ -55,9 +55,6 @@ use alloc::boxed::Box;
 use alloc::vec::Vec;
 
 #[cfg(feature = "std")]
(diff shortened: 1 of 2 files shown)


bkchr:
The `no_std` error is implemented for every type and does not need to be re-exported. Given that, I close this.

PR #34, opened 2019-11-09
Implement OwnedNode using a node decode plan.

jimpo:
The goal of this change is to remove some of the duplication and inefficiency of `OwnedNode`. `OwnedNode` is a parallel data structure to `Node` that is constructed by copying a lot of data into new allocations. Notably, an `OwnedNode` cannot easily be converted to a `Node` because of alignment issues between `NibbleVec`, which `OwnedNode` uses to represent partial keys, and `NibbleSlice`, which `Node` uses to represent partial keys.

The idea head is for `OwnedNode` to just store an encoded trie node along with a blueprint for quickly decoding into a node. When an encoded node is first decoded, instead of creating a `Node` directly, which has a lifetime, one can construct a `NodePlan` which is an owned structure and can be used to quickly create Nodes thereafter. The `OwnedNode` can be cheaply converted into a `Node` by using the `NodePlan` on its owned data.

*This is a breaking change as it modifies the `NodeCodec` trait and adds a new required method. This probably needs a minor version bump at a minimum.* The actual code changes required to upgrade are fairly minimal though.

Using the newly added benchmark, trie traversal on an in-memory trie is 42% faster.

### DIFF
--- a/test-support/reference-trie/src/lib.rs
+++ b/test-support/reference-trie/src/lib.rs
@@ -21,7 +21,7 @@ use std::ops::Range;
 use parity_scale_codec::{Decode, Input, Output, Encode, Compact, Error as CodecError};
 use trie_root::Hasher;
 use trie_db::{
-	node::{NibbleSlicePlan, NodePlan, Node},
+	node::{NibbleSlicePlan, NodePlan},
 	triedbmut::ChildReference,
 	DBValue,
 	trie_visit,
@@ -430,26 +430,6 @@ fn decode_size<I: Input>(first: u8, input: &mut I) -> Result<usize, CodecError> 
 		result += 255;
 	}
 	Err("Size limit reached for a nibble slice".into())
-}
-
-#[test]
-fn test_encoding_simple_trie() {
-	for prefix in [
-		LEAF_PREFIX_MASK_NO_EXT,
-		BRANCH_WITHOUT_MASK_NO_EXT,
-		BRANCH_WITH_MASK_NO_EXT,
-	].iter() {
-		for i in (0..1000).chain(NIBBLE_SIZE_BOUND_NO_EXT - 2..NIBBLE_SIZE_BOUND_NO_EXT + 2) {
-			let mut output = Vec::new();
-			encode_size_and_prefix(i, *prefix, &mut output);
-			let input	= &mut &output[..];
-			let first = input.read_byte().unwrap();
-			assert_eq!(first & (0b11 << 6), *prefix);
-			let v = decode_size(first, input);
-			assert_eq!(Ok(std::cmp::min(i, NIBBLE_SIZE_BOUND_NO_EXT)), v);
-		}
-
-	}
 }
 
 impl Encode for NodeHeaderNoExt {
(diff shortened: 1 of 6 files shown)


test-support/reference-trie/src/lib.rs
```diff
@@ -585,6 +557,57 @@ fn partial_encode(partial: Partial, node_kind: NodeKindNoExt) -> Vec<u8> {
 	output
 }
 
+struct ByteSliceInput<'a> {
```
cheme: - take method now uses a struct (butesliceinput)
  -> do we want to move this bytesliceinput in a common crate: take was already probably copied between implementations, but this is a bit more code.
  I would suggest to put it nearby the Codec, and NodePlan place.
- Or could we use std 'Cursor' instead of bytesliceinput.
- decode plan could use 'cursor', 'byteslice' or 'read' trait implementation as a parameter.
 Read trait is my favorite as it can extend to nocopy proof decoding.


- Or could we use std 'Cursor' instead of bytesliceinput.

trie-db/src/iterator.rs
```diff
@@ -77,30 +76,31 @@ impl<'a, L: TrieLayout> TrieDBNodeIterator<'a, L> {
 			trail: Vec::with_capacity(8),
 			key_nibbles: NibbleVec::new(),
 		};
-		db.root_data().and_then(|root_data| {
-			r.descend(&root_data, Some(db.root().clone())).map(|_| ())
-		})?;
+		let (root_node, root_hash) = db.get_raw_or_lookup(db.root().as_ref(), EMPTY_PREFIX)?;
```
cheme: - root_data become root function, there is only one place root_data seems still use: maybe lets remove the function or keep using it but not a middle ground.
(iterator.rs line 206).

trie-db/src/iterator.rs
```diff
@@ -77,30 +76,31 @@ impl<'a, L: TrieLayout> TrieDBNodeIterator<'a, L> {
...
+			let (next_node, next_node_hash) = {
+				self.descend(node, node_hash);
 				let crumb = self.trail.last_mut()
 					.expect(
 						"descend_into_node pushes a crumb onto the trial; \
 						thus the trail is non-empty; qed"
 					);
 
-				match node {
+				match crumb.node.node() {
```
cheme: To get performance improvement, the number of 'node' call would probably need to be reduce, or maybe use OwnedNode<&'a[u8]> in crumb (not really sure).
Owned<&'a[u8]> for zero copy proof check, is also a bit of a challenge regarding lifetime management.

trie-db/src/triedb.rs
```diff
@@ -298,16 +297,15 @@ impl<'a, L: TrieLayout> Iterator for TrieDBIterator<'a, L> {
 		while let Some(item) = self.inner.next() {
 			match item {
 				Ok((mut prefix, _, node)) => {
-					let maybe_value = match node.as_ref() {
-						&OwnedNode::Leaf(ref partial, ref value) => {
-						prefix.append(partial);
-							Some(value.as_ref())
+					let maybe_value = match node.node() {
+						Node::Leaf(partial, value) => {
+							prefix.append_partial(partial.right());
```
cheme: switch to append_partial is an additional cost here, the method is 'slow', I don't know how much it can be responsible for perf not being as expected.

trie-db/src/node.rs
```diff
@@ -41,88 +42,112 @@ pub enum Node<'a> {
...
+	Extension {
+		partial: Range<usize>,
+		partial_padding: usize,
+		child_data: Range<usize>,
+	},
+	/// Branch node; has slice of child nodes (each possibly null)
+	/// and an optional immediate node data.
+	Branch {
+		value: Option<Range<usize>>,
+		children: [Option<Range<usize>>; nibble_ops::NIBBLE_LENGTH],
```
cheme: we are using a bit useless memory here: can be twice as small approximately, just putting index into slice, this seems to favor code readability over optimization that does not really make a difference. I am not sure where to stand here.

trie-db/src/node.rs
```diff
@@ -41,88 +42,112 @@ pub enum Node<'a> {
...
-	Empty,
-	/// Leaf node: partial key and value.
-	Leaf(NibbleVec, DBValue),
-	/// Extension node: partial key and child node.
-	Extension(NibbleVec, DBValue),
-	/// Branch node: children and an optional value.
-	Branch(Branch),
-	/// Branch node: children and an optional value.
-	NibbledBranch(NibbleVec, Branch),
+pub struct OwnedNode<D: AsRef<[u8]>> {
```
cheme: - 'OwnedNode' naming is confusing to me as it can now be 'OwnedNode<&'a[u8]>'.

trie-db/src/node.rs
jimpo: Well, in this case all the memory is on the stack whereas a more memory-efficient alternative would require a dynamically sized Vec, requiring a heap allocation. This also automatically ensures that none of the indices are out of range. On the other hand, arrays are kind of a pain to work with since they are not `IntoIterator`. IMO, it's not worth changing though.

trie-db/src/triedb.rs
jimpo: I don't think this is any slower than the `append` it was doing before.

trie-db/src/node.rs
jimpo: Changed from `D: AsRef<[u8]>` to `D: Borrow<[u8]>`.

trie-db/src/iterator.rs
jimpo: Refactored this method to significantly reduce the number of calls to `node()`.

test-support/reference-trie/src/lib.rs
jimpo: Definitely there's too much duplication between `NodeCodec` implementations, but I'd rather leave the task of dealing with that and refactoring into utilities for another PR.

jimpo:
@cheme Thanks for pointing out the regression.

I added a benchmark and some performance optimizations. I am now seeing at 18.8% perf improvement in this PR over master for trie iteration.

trie-db/src/triedb.rs
cheme: If I trust the comment it is shifting nibbles when vec is always aligned.

test-support/reference-trie/src/lib.rs
cheme: Do you think 'Cursor' could be use instead? Iirc you could not yes mut ref over slice because it could not go backward, but set_position or seek should be fine.

jimpo: I think it could work, but `Cursor` is in `std::io` which is not imported into the Substrate runtime with `no_std`. And we will want the trie codecs to work inside the runtime. Also, `ByteSliceInput` is a pretty small amount of code IMO.

trie-db/src/triedb.rs
jimpo: Yes, this is true. But this just changes from `NibbleVec::append` to `NibbleVec::append_partial` and `NibbleVec::append` has exactly the same performance issue. It's just that the parameter type is different.

test-support/reference-trie/src/lib.rs
cheme: Oh yes, I ovelooked that, Read, Write and Cursor should move to alloc crate in my opinion (afaik only box is responsible for it being in io).

trie-db/src/node.rs
cheme: I was just suggesting to store only one index per children. But it relies on assumption that each children are separated by a constant number of byte (1 in substrate).
But such assumption make codec trait less practical, it is not a good idea.

trie-db/benches/bench.rs
```diff
@@ -445,3 +446,20 @@ fn trie_mut_build_b(c: &mut Criterion) {
...
+
+fn trie_iteration(c: &mut Criterion) {
+	use memory_db::HashKey;
+
+	let input = input2(29, 204800, 32);
+
+	let mut mdb = memory_db::MemoryDB::<_, HashKey<_>, _>::default();
+	let root = reference_trie::calc_root_build(input, &mut mdb);
+
+	c.bench_function("quick_iter_bench", move |b: &mut Bencher|
```
cheme:
```suggestion
	c.bench_function("trie_iteration", move |b: &mut Bencher|
```

trie-db/src/iterator.rs
```diff
@@ -184,94 +201,85 @@ impl<'a, L: TrieLayout> TrieDBNodeIterator<'a, L> {
...
-						} else {
-							IterStep::Continue
+				let b = self.trail.last_mut()?;
+
+				if let Status::Entering = b.status {
+					// Handle the Status::Entering case separately to avoid constructing the node
+					// unnecessarily.
+					IterStep::YieldNode
+				} else {
+					let node = b.node.node();
```
cheme: naïve question, can't we just match on NodePlan instead, and only access the fields if they are needed.

jimpo: Nice, didn't think of that! With this, it's a 43% speedup over master!

PR #35, opened 2019-11-11
Add NodeHandle enum for node references within Node

jimpo:
Currently, branch and extension nodes just have a byte slice reference to the child nodes and one uses the `NodeCodec::try_decode_hash` method to determine whether it is a hash or an inline node reference. Instead, we make this an explicit part of the decoded Node structure to simply code and make the codec more flexible. For example, we can stop encoding the length of the hashes inside branch nodes and instead use another bitfield inside branch nodes to indicate whether each child is a hash or inline reference, which should save space.

This is another breaking change to NodeCodec building on #34.

### DIFF
--- a/trie-db/src/triedbmut.rs
+++ b/trie-db/src/triedbmut.rs
@@ -106,8 +106,8 @@ where
 		storage: &mut NodeStorage<H::Out>
 	) -> Result<NodeHandle<H::Out>, H::Out, C::Error>
 	where
-		C: NodeCodec<H>,
-		H: Hasher<Out = O>,
+		C: NodeCodec<HashOut=O>,
+		H: Hasher<Out=O>,
 	{
 		let handle = match child {
 			EncodedNodeHandle::Hash(data) => {
@@ -131,7 +131,7 @@ where
 		storage: &'b mut NodeStorage<H::Out>,
 	) -> Result<Self, H::Out, C::Error>
 		where
-			C: NodeCodec<H>, H: Hasher<Out = O>,
+			C: NodeCodec<HashOut = O>, H: Hasher<Out = O>,
 	{
 		let encoded_node = C::decode(data)
 			.map_err(|e| Box::new(TrieError::DecoderError(node_hash, e)))?;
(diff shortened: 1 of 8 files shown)


trie-db/src/triedbmut.rs
```diff
@@ -100,72 +100,84 @@ where
 {
 	// load an inline node into memory or get the hash to do the lookup later.
 	fn inline_or_hash<C, H>(
-		node: &[u8],
+		parent_hash: H::Out,
```
cheme: 'parent_hash' as a pointer here?

trie-db/src/triedbmut.rs
```diff
@@ -100,72 +100,84 @@ where
...
-			.unwrap_or_else(|| {
-				let child = Node::from_encoded::<C, H>(node, db, storage);
+		let handle = match child {
+			EncodedNodeHandle::Hash(data) => {
+				let hash = decode_hash::<H>(data)
+					.ok_or_else(|| Box::new(TrieError::InvalidHash(parent_hash, data.to_vec())))?;
+				NodeHandle::Hash(hash)
+			},
+			EncodedNodeHandle::Inline(data) => {
+				let child = Node::from_encoded::<C, H>(parent_hash, data, db, storage)?;
```
cheme: is it possible that the small perf hit comes from getting a decoded data handle when previously it was encoded data? I don't really think so (exept if from branch this method is called on child we do not need, but it should not), but I am not sure where it comes from.

trie-db/src/triedbmut.rs
```diff
@@ -100,72 +100,84 @@ where
...
+						.map(Some),
+					None => Ok(None),
 				};
 
 				let children = Box::new([
-					child(0), child(1), child(2), child(3),
-					child(4), child(5), child(6), child(7),
-					child(8), child(9), child(10), child(11),
-					child(12), child(13), child(14), child(15),
+					child(0)?, child(1)?, child(2)?, child(3)?,
```
cheme: ok we are doing it here.

cheme:
No sorry, I misread the code, there is no such static rules.

arkpar:
Needs a rebase

arkpar:
Performance hit might come from `from_encoded` returning `Result` now. I think proper error handling is still worth it though.

trie-db/src/triedbmut.rs
cheme: I believe the performance hit may come for the fact that before this pr, call to child here only went into try_decode_hash that do not do inline node decoding.
After this pr for all inline node (even if not needed) we run a decoding step.

cheme: But it shouldn't matter for iteration since all child are queried, for other operation it can, but it is probably minor.

PR #36, opened 2019-11-16
Disable "ahash-compile-time-rng" feature in hashbrown.

eddyb:
This will ensure that every build uses the same `ahash` seed, instead of a compile-time randomized one (which is now the default - see https://github.com/rust-lang/hashbrown/issues/124, https://github.com/rust-lang/hashbrown/pull/125, https://github.com/tkaitchuck/aHash/pull/18 and https://github.com/tkaitchuck/aHash/pull/25 for more details/background).

(The other such update PR needed is https://github.com/paritytech/finality-grandpa/pull/90).

### DIFF
--- a/memory-db/Cargo.toml
+++ b/memory-db/Cargo.toml
@@ -10,7 +10,7 @@ license = "Apache-2.0"
 heapsize = { version = "0.4", optional = true }
 parity-util-mem = { version = "0.2", default-features = false }
 hash-db = { path = "../hash-db", default-features = false, version = "0.15.2"}
-hashbrown = { version = "0.6.0" }
+hashbrown = { version = "0.6.3", default-features = false, features = [ "ahash" ] }
 
 [dev-dependencies]
 keccak-hasher = { path = "../test-support/keccak-hasher", version = "0.15.2"}
--- a/trie-db/Cargo.toml
+++ b/trie-db/Cargo.toml
@@ -11,7 +11,7 @@ log = "0.4"
 rand = { version = "0.6", default-features = false }
 elastic-array = { version = "0.10", default-features = false }
 hash-db = { path = "../hash-db", default-features = false, version = "0.15.2"}
-hashbrown = { version = "0.6" }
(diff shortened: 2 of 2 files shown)


eddyb:
[clabot:check]

arkpar:
Does it interfere with `no_std` build? Randomized seed is extra security against hash collision attacks.  And I don't think we require consistent ordering anywhere, as we switched from std to hashbrown not that long ago. I'd keep it as is at least for `std`

eddyb:
It depends if you want the build process (and the resulting binaries) to be deterministic (with this change) or not (without) - cc @rphmeier who merged https://github.com/paritytech/finality-grandpa/pull/90.

One thing I'll note is that any published binaries (e.g. WASM blobs on-chain) will have the seed extractable from them anyway, so that should be included in the attack model for hash collisions.

burdges:
We've that DoS concern but also the opposite worry @arkpar 

It appears `MemoryDB::drain` returns a `hashbrown::HashMap` which provides `Iterator` methods.  If this iteration order ever impacts final behavior anywhere then attackers could make nodes behave differently, possibly increasing the attackers' power, either by confusion or maybe by slashing and disabling many nodes

I'd suggest tools like https://docs.rs/ordermap/0.4.2/ordermap/ anytime we need this iteration, but maybe `drain` could just be replaced by a `clear` method that returned nothing. 

We could construct random seeds from the chain state so that everyone had equal seeds right now, like a Fiat-Shamir transform, except not sure if this really helps the DoS case.

rphmeier:
> One thing I'll note is that any published binaries (e.g. WASM blobs on-chain) will have the seed extractable from them anyway, so that should be included in the attack model for hash collisions.

This was my reasoning for merging the PR against finality-grandpa. Randomizing the seed only matters if you do so on each run of the program. In the case of on-chain WASM, that is not the case. That bears re-evaluation of any of our usages of hash maps in our no-std crates, but that's beyond the scope of the PR.

PR #42, opened 2019-11-27
Create TrieDBBrokenIterator replicating version <0.16 iterator bug.

jimpo:
In versions <0.16.0 of trie-db, a bug in `TrieDBIterator` would cause `seek` to position the cursor incorrectly in some cases. In particular, the cursor would be positioned at the beginning of a branch or extension node even if the key parameter follows it. This bug has become part of the consensus logic on certain deployed blockchains including Kusama.

This is part of the solution for fixing the storage root mismatch bug on Kusama.


jimpo:
Instead, a Kusama CC3 was released, so backwards compatibility is not required.

PR #44, opened 2019-12-05
Implement compact encoding scheme for partial tries.

jimpo:
As per module doc:

```rust
//! Compact encoding/decoding functions for partial Merkle-Patricia tries.
//!
//! A partial trie is a subset of the nodes in a complete trie, which can still be used to
//! perform authenticated lookups on a subset of keys. A naive encoding is the set of encoded nodes
//! in the partial trie. This, however, includes redundant hashes of other nodes in the partial
//! trie which could be computed directly. The compact encoding strips out all hash child
//! references to other nodes in the partial trie and replaces them with empty inline references,
//! indicating that the child reference is omitted. The nodes are then ordered in pre-order
//! traversal order so that the full nodes can be efficiently reconstructed recursively. Note that
//! hash references to nodes not in the partial trie are left intact. The compact encoding can be
//! expected to save roughly (n - 1) hashes in size where n is the number of nodes in the partial
//! trie.
```

### DIFF
--- a/trie-db/src/trie_codec.rs
+++ b/trie-db/src/trie_codec.rs
@@ -245,7 +245,8 @@ pub fn encode_compact<L>(db: &TrieDB<L>) -> Result<Vec<Vec<u8>>, TrieHash<L>, CE
 			}
 			Err(err) => match *err {
 				// If we hit an IncompleteDatabaseError, just ignore it and continue encoding the
-				// incomplete trie.
+				// incomplete trie. This encoding must support partial tries, which can be used for
+				// space-efficient storage proofs.
 				TrieError::IncompleteDatabase(_) => {},
 				_ => return Err(err),
 			}
--- a/trie-db/src/trie_codec.rs
+++ b/trie-db/src/trie_codec.rs
@@ -1,0 +1,635 @@
+// Copyright 2019 Parity Technologies
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Compact encoding/decoding functions for partial Merkle-Patricia tries.
+//!
+//! A partial trie is a subset of the nodes in a complete trie, which can still be used to
+//! perform authenticated lookups on a subset of keys. A naive encoding is the set of encoded nodes
+//! in the partial trie. This, however, includes redundant hashes of other nodes in the partial
+//! trie which could be computed directly. The compact encoding strips out all hash child
+//! references to other nodes in the partial trie and replaces them with empty inline references,
+//! indicating that the child reference is omitted. The nodes are then ordered in pre-order
+//! traversal order so that the full nodes can be efficiently reconstructed recursively. Note that
+//! hash references to nodes not in the partial trie are left intact. The compact encoding can be
+//! expected to save roughly (n - 1) hashes in size where n is the number of nodes in the partial
+//! trie.
+
+use hash_db::HashDB;
+use crate::{
+	CError, ChildReference, DBValue, NibbleVec, NodeCodec, Result,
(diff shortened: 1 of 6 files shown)


trie-db/src/trie_codec.rs
```diff
@@ -0,0 +1,635 @@
...
+	prefix: NibbleVec,
+	node: Rc<OwnedNode<DBValue>>,
+	/// The next entry in the stack is a child of the preceding entry at this index. For branch
+	/// nodes, the index is in [0, NIBBLE_LENGTH] and for extension nodes, the index is in [0, 1].
+	child_index: usize,
+	/// Flags indicating whether each child is omitted in the encoded node.
+	omit_children: Vec<bool>,
+	/// The encoding of the subtrie nodes rooted at this entry, which is built up in
+	/// `encode_compact`.
+	encoded_nodes: Vec<Vec<u8>>,
```
cheme: I am not sure why it is in Entry: shouldn't it be outside this struct ?
Like using a single VecDeque and avoid those multiple vecs instance.

trie-db/src/trie_codec.rs
```diff
@@ -0,0 +1,635 @@
...
+		match self.node.node_plan() {
+			NodePlan::Empty | NodePlan::Leaf { .. } =>
+				return Err("empty and leaf nodes have no children"),
+			NodePlan::Extension { .. } => {
+				if self.child_index != 0 {
+					return Err("extension node cannot have multiple children")
+				}
+			}
+			NodePlan::Branch { .. } => {
+				if child_prefix.len() <= self.prefix.len() {
```
cheme: I wonder if we could just debug_assert that (the function is only call with expect).

trie-db/src/trie_codec.rs
```diff
@@ -0,0 +1,635 @@
...
+					return Err("child_prefix does not contain prefix");
+				}
+				let child_index = child_prefix.at(self.prefix.len()) as usize;
+				if child_index < self.child_index {
```
cheme: same question on just debug asserting it, also isn't 'child_index == self.child_index' another thing that should not happen?

trie-db/src/trie_codec.rs
```diff
@@ -0,0 +1,635 @@
...
+			}
+		}
+		Ok(())
+	}
+
+	/// Marks a child entry as the child at the current index.
+	fn omit_child(&mut self, child_entry: Self)
+		-> Result<(), C::HashOut, C::Error>
+	{
+		self.encoded_nodes.extend(child_entry.encode_subtrie()?);
```
cheme: encoded_nodes as a single parameter (&mut VecDeque)?

trie-db/src/trie_codec.rs
```diff
@@ -0,0 +1,635 @@
...
+					node,
+					child_index: 0,
+					omit_children: vec![false; children_len],
+					encoded_nodes: Vec::new(),
+					_marker: PhantomData::default(),
+				});
+			}
+			Err(err) => match *err {
+				// If we hit an IncompleteDatabaseError, just ignore it and continue encoding the
+				// incomplete trie.
```
cheme: A quick additional word on the fact that we use it to process proof, and that for this use case the trie is inherently incomplete (to get the rational behind this skip).

trie-db/src/triedbmut.rs
```diff
@@ -282,11 +284,39 @@ enum Stored<H> {
 }
 
 /// Used to build a collection of child nodes from a collection of `NodeHandle`s
-pub enum ChildReference<HO> { // `HO` is e.g. `H256`, i.e. the output of a `Hasher`
+#[derive(Clone, Copy)]
+pub enum ChildReference<HO: Copy> { // `HO` is e.g. `H256`, i.e. the output of a `Hasher`
```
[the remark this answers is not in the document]
cheme: seems like tests are passing without the copy three additional copy bound of this pr, but it may be here for something I did not see.

trie-db/src/trie_codec.rs
jimpo: It's OK if they are equal. After a child reference is populated, the index is incremented.

trie-db/src/triedbmut.rs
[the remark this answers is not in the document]
jimpo: Yes, you are right. I forgot that the `derive(Copy)` happens opportunistically even without the trait bound.

trie-db/src/trie_codec.rs
jimpo: I figured out how to get rid of this.

PR #46, opened 2019-12-12
Replace ElasticArray with SmallVec

dvdplm:
The `elastic-array` crate is a fine piece of software but it'd be good to prefer crates that are commonly used in the ecosystem and likely to be well maintained. `smallvec` is at `v1.0` and generally well-regarded.

Unfortunately as `DBValue` is part of the public API of `trie-db`, this is a breaking change so changing this implies annoying busy-work in consuming code.

### DIFF
--- a/trie-db/src/nibble/mod.rs
+++ b/trie-db/src/nibble/mod.rs
@@ -16,7 +16,7 @@
 
 mod nibblevec;
 mod nibbleslice;
-use elastic_array::ElasticArray36;
+use smallvec::SmallVec;
 use crate::node::NodeKey;
 use core_::cmp;
 
@@ -152,7 +152,7 @@ pub mod nibble_ops {
 #[cfg_attr(feature = "std", derive(Debug))]
 #[derive(Clone, PartialEq, Eq)]
 pub struct NibbleVec {
-	inner: ElasticArray36<u8>,
+	inner: SmallVec<[u8; 36]>,
 	len: usize,
 }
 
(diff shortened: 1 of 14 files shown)


trie-db/src/nibble/mod.rs
```diff
@@ -152,7 +152,7 @@ pub mod nibble_ops {
 #[cfg_attr(feature = "std", derive(Debug))]
 #[derive(Clone, PartialEq, Eq)]
 pub struct NibbleVec {
-	inner: ElasticArray36<u8>,
+	inner: SmallVec<[u8; 36]>,
```
jimpo: This type is used a lot and the hardcoded 36 kind of bothers me. Could you typedef with `pub type ByteVec = SmallVec<[u8; 36]>`.

Naming-wise, I feel like the name `nibble::ByteVec` is probably fine or maybe `BackingVec` or something like that.

dvdplm: Yeah, I was on the fence on that, agree it's noisy. 
I did something similar to what you suggest [here](https://github.com/paritytech/parity-common/pull/282/files#diff-7fe73f2989f300cb5d5dc23a9a54a212R31). Reason I had doubts is that the type sort of seeps through and ends up everywhere, so maybe it's better to be explicit about it?
Happy to change this though.

jimpo: Yeah, I'd prefer the typedef. And the `DBKey` type makes sense to me as well.

dvdplm:
> I think it is worth considering making DBValue Vec (in many case it ends up being converted to Vec).

I went ahead and implemented this in [b18e2db](https://github.com/paritytech/trie/pull/46/commits/b18e2db667a400718d5e3b0a2706c46532768d94) and the terrible perf regression seems to have gone away. A bit unsure of how big the impact is on the substrate side though.

PR #47, opened 2019-12-17
Prepare versioning for 0.18 crates (no elastic array)

cheme:
This is versioning change for elastic array removal, (I did not include #39 (only one review) but it can be release in a later patch version).

**It still requires that new version of parity-util-mem get published first (see TODO in memorydb Cargo.toml).**

Note that it seems like some crate do not have to move to 0.18 (because hashdb is untouched from this change).

### DIFF
--- a/memory-db/Cargo.toml
+++ b/memory-db/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "memory-db"
-version = "0.15.3"
+version = "0.18.0"
 authors = ["Parity Technologies <admin@parity.io>"]
 description = "In-memory implementation of hash-db, useful for tests"
 repository = "https://github.com/paritytech/trie"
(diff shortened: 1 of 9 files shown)


memory-db/Cargo.toml
```diff
@@ -1,14 +1,14 @@
...
+version = "0.18.0"
 authors = ["Parity Technologies <admin@parity.io>"]
 description = "In-memory implementation of hash-db, useful for tests"
 repository = "https://github.com/paritytech/trie"
 license = "Apache-2.0"
 
 [dependencies]
 heapsize = { version = "0.4", optional = true }
-parity-util-mem = { version = "0.2", default-features = false }
+parity-util-mem = { version = "0.2", default-features = false } # TODO this requires update to latest version
```
dvdplm: This will be `0.3`

dvdplm:
Moved over to https://github.com/paritytech/trie/pull/48

PR #50, opened 2019-12-23
Fix wasm compilation of 0.18

NikVolf:
See
https://gitlab.parity.io/parity/substrate/-/jobs/338788

There should be also CI check for that


NikVolf:
required for https://github.com/paritytech/substrate/pull/4483

PR #56, opened 2019-12-30
fix for memory-db-bench (error[E0373])

jsulmont:
This PR allows `trie` to compile with recent nightly.


jsulmont:
[clabot:check]

PR #77, opened 2020-02-06
Allow constant associated empty root value.

cheme:
This PR adds new trait `HasherNullEmptyRoot` that keep trace of an associated null value for an empty type.
A new trait is used because this only makes sense if we assumes that the codec use in the trie will encode node to [0] (this is a bit fishy but rather straightforward solution).
Another memory db is implemented that uses this trait to avoid having to calculate the empty trie root (this happens a lot in substrate).
The implementation is generated from a macro and I am not very happy with the result, but again did not find any good and simple way to implement that, suggestions are very welcome.
Not that in the macro I did ommit an indentation on purpose to keep the diff readable, it should be added before merging.

### DIFF
--- a/hash-db/src/lib.rs
+++ b/hash-db/src/lib.rs
@@ -64,6 +64,21 @@ pub trait Hasher: Sync + Send {
 	fn hash(x: &[u8]) -> Self::Out;
 }
 
+/// Technical trait to avoid calculating empty root.
+/// This trait assumes, an empty node is `[0u8]`.
+pub trait HasherNullEmptyRoot: Hasher {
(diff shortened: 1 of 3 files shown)


burdges:
Is this for domain separation in the hash?

cheme:
I am not sure about what you mean by "domain separation in the hash", but I it is probably not, it is just to avoid this kind of code https://github.com/paritytech/substrate/blob/faf608eeeda0031629f7608321c7ab41fac6024f/primitives/trie/src/lib.rs#L217 in substrate.

PR #126, opened 2021-01-29
Allow skipping value for compact proof

cheme:
This PR extend trie_codec with the ability to selectively remove the values from the compact proof.

This differs from trie_db/src/proof by working on any proofs.

### Encoding (compacting)

Values are removed by applying a given FnMut condition.

Values are replaced by the 0 length value.

We can escape actual 0 length value if needed, it is not needed when:
- trie layout that does not allow empty value
- trie that only contain value in leaf (constant size key) likely don't need it
either as 0 length value will be a inline node which cannot be removed.
- the proof is only decoded given a set of known value (we pass the value that can
be replaced as parameter).

### Decoding

Two possibilities:
- we provide the keys that have been removed (we can provide more).
- we try to replace every empty values in the proof


burdges:
Are you removing values known due to all leaves being present?  Or some other reason makes them redundent?

cheme:
No, the values are removed when they are known. eg you check some proof, then keep some of the value and use a different proof for which you know those value can be use.

So it is just to remove values already known by the verifier.

In practice the current target is to remove the ':code' key value from cumulus collated proof because with current cumulus we have the validation wasm being the same as the runtime wasm (requires a new polkadot host method to fetch this wasm lazilly when needed by the proof).
I made the code a bit more flexible (remove with any condition, inject multiple keys).

burdges:
It's for storage then?   I see.. 

Yes, we definitely need code to be handled separately.  lol

cheme:
Forgot to link a 'rough' (everything at the wrong place) implementation https://github.com/paritytech/cumulus/pull/295 of the :code case.
Also food for thought, on putting value aside (or optionally attached) and allow some optional caching. Generally having something read by runtime at every block (and not written) is something that could be optimized out of the validation proof.

PR #127, opened 2021-03-01
Seek for prefixed iterator.

cheme:
When using 'seek' on a prefixed iterator, prefix context is lost.

This pr adds a function 'prefix_then_seek' for this usecase.


cheme:
CI is blocked behind https://github.com/bitvecto-rs/bitvec/pull/110 (see https://github.com/paritytech/parity-scale-codec/pull/253 for context).

arkpar (approved):

PR #134, opened 2021-05-28
Inner hash of value support and metadata storage.

cheme:
This PR allows Inner hash of value support and metadata storage.

# Inner hash of value

The triedbmut code change to allow hash of value instead of value:
enum `Value` replaces `Option<DBValue>`.

When accessing hash, then `HashDB` trait allows querying for value through
`access_from` hashdb new function, but at this point it only acts as
callback of value access (since in all hashdb implementation values kept
being stored with their encoded node).

Access from call is very relevant also to mark access to value when building
proofs (through transiant meta).

# Metadata storage

Additional metadata is added to nodes.
Meta content can be transiant or stored or even part of state (encoded node).

## Stored with node

These metadata will not change root hash, 

Depending on usage it can indirectly modify it.
eg: choice of hash implementation depending on trie current layout.
This is only ok if the trie do support both behavior, but should generally be
avoided (by storing some metadata in state for instance).

But generally it should be related to technical implementation needs.

This works by modifying `HashDB` to allow storing metadata with encoded node,
new function `get_with_meta` and instert_with_meta` should be use.

## Stored in node state

These metadata will change root hash and are part of state.

For this we modified codec to use meta as input and also update meta after decoding.
(also allows codec to use different encoding depending on meta (not use at this point
but use in current target but use in other branch where new encoding was totally new).

`read_meta` and `write_meta` are used for it.

## TODO

major versioning of crates (hashdb change means that pushing all +0.1.0 is probably the easiest).


cheme:
This PR will be refactor to use a different encoding scheme to remove most use of additional metadata.

trie-db/src/triedbmut.rs
```diff
@@ -1625,11 +1776,4 @@ mod tests {
 		test_comb((0, &a), (1, &b), (1, &[0x01, 0x23, 0x46, 0x78][..]));
 		test_comb((1, &a), (1, &b), (0, &[0x23, 0x46, 0x78][..]));
 	}
-
-	#[test]
-	fn nice_debug_for_node() {
```
cheme, line 1630: Removed due to absence of an easy way to have a Layout with a u32 hash.
(could switch back to having Node<H> instead of Node<L>, but switching to Node<L> is something I did a few time in the past so I did not revert those changes, but doable).

cheme:
To reviewers, looking at current changes, 'Meta' in trielayout seems a bit awkward, I therefore recreate a branch with a inner meta type:
https://github.com/cheme/trie/compare/hashed-value-simple4...cheme:hashed-value-simple4-meta-trait

This is more changes, but I generally find the typing better.
So opinion on this could be interesting (I will maintain both branches).

arkpar:
After an internal discussion we've agreed to try a different approach. Values will be stored as plain data in `HashDb` and not embedded in value nodes. This will make implementation much more simple as no special handling in `HashDb` would be required.

qdm12:
Hello everyone, I'm just curious if this PR is still relevant? 🤔 Thanks!!

cheme:
No, it was superseded by #142, thanks for noticing it.

PR #142, opened 2021-09-03
Inner hash of value with new nodes

cheme:
This PR aims at replacing #134.
It achieves the same goals but always use a new db node when inner hashing is use.
This makes code simpler and can be interesting with rc dbs.

Note that compact proof and compact proof with value iterator did change a bit:
- compact proof: hashed value are following the node (node is switch to a standard node with 0 length value), which is prefixed with a header dead value.
- compact proof with value iter: hashed value when ignored are replace similarily, just no value node are added afterward.

At this point review can be started as all remaining TODOs are minor.
Codec in substrate_like.rs is the same as the one which will be ported in substrate except for the default threshold size (set to 1 here as most of test only use small values).

### TODOs
- crate versioning
- write a test for value removal with prefix (check the value node prefix is correct) + a test that insert twice in triedbmut
- https://github.com/paritytech/trie/pull/142/commits/63e1651b7647be6267e6da39ec0ea06d5b7c466f is related to a possible migration optimisation in state machine, does not really have to be in this PR.

- consider running rust format (little care was used here)


trie-db/src/triedbmut.rs
```diff
@@ -415,13 +517,15 @@ pub struct TrieDBMut<'a, L>
...
-	storage: NodeStorage<TrieHash<L>>,
+	layout: L,
+	storage: NodeStorage<L>,
 	db: &'a mut dyn HashDB<L::Hash, DBValue>,
 	root: &'a mut TrieHash<L>,
 	root_handle: NodeHandle<TrieHash<L>>,
 	death_row: HashSet<(TrieHash<L>, (BackingByteVec, Option<u8>))>,
 	/// The number of hash operations this trie has performed.
 	/// Note that none are performed until changes are committed.
+	/// This omit value hashing.
```
arkpar, line 528:
```suggestion
	/// This does not include value hashing.
```

trie-db/src/triedbmut.rs
```diff
@@ -431,10 +535,21 @@ where
 {
 	/// Create a new trie with backing database `db` and empty `root`.
 	pub fn new(db: &'a mut dyn HashDB<L::Hash, DBValue>, root: &'a mut TrieHash<L>) -> Self {
+		Self::new_with_layout(db, root, Default::default())
+	}
+
+	/// Create a new trie with backing database `db` and empty `root`.
+	/// This could use a context specific layout.
```
arkpar, line 542:
```suggestion
	/// using a context specific layout.
```

arkpar (changes requested):
I'd rename "AltHash" that is used everywhere to something else. It implies that an alternative hashing scheme or an alternative hasher is used. While in fact this is just an alternative trie structure. Maybe something like "ExtValue"?

What's the use case for `OmitValue` and `AttachNode` proof modes?

cheme:
> I'd rename "AltHash" that is used everywhere to something else. It implies that an alternative hashing scheme or an alternative hasher is used. While in fact this is just an alternative trie structure. Maybe something like "ExtValue"?

Indeed, does not make really sense now, I thing I will use 'HashedValue' as it is already the name of a variant.

> What's the use case for `OmitValue` and `AttachNode` proof modes?

When compacting proof, 'OmitValue' is the previous state where the value is removed from proof because it is attached to the iterator, 'AttachNode' is the same when the value is hash of the node:
I should rename it 'OmitHashedValue'.
This is needed to know if the value from iterator on verification needs to be hashed or not. That is needed as long as the state can be hybrid (if not we can pass the hash threshold as verifier parameter and deduce it from value length).

trie-db/src/triedbmut.rs
```diff
@@ -66,23 +67,117 @@ fn empty_children<H>() -> Box<[Option<NodeHandle<H>>; 16]> {
...
 
+/// Value representation for Node.
+#[derive(Clone, Eq)]
+pub enum Value<L: TrieLayout> {
+	/// Node with no value attached.
+	NoValue,
+	/// Value bytes.
+	Value(DBValue),
+	/// Hash of value bytes if calculated and value bytes.
+	HashedValue(Option<TrieHash<L>>, Option<DBValue>),
```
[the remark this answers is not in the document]
cheme: It is None - None that is not possible.
None - Some is a new value (trie hash calculated later).
Some - None is a node read but value unaccessed.
Some - Some is a node read and value accessed.
So I can switch to:
HashedValueUnresolved(ThieHash<L>)
and
NewHashedValue(Option<TrieHash<L>>, DBValue)
(or three variant, but these two seems enough)

trie-db/src/node.rs
```diff
@@ -108,6 +140,30 @@ impl NibbleSlicePlan {
 	}
 }
 
+/// Plan for value representation in `NodePlan`.
+#[derive(Eq, PartialEq, Clone)]
+#[cfg_attr(feature = "std", derive(Debug))]
+pub enum ValuePlan {
```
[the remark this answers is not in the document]
cheme: Also when value is unaccessed (this is the part of the node plan that reflect if the header was the variant with value hashing or not).

trie-db/src/proof/generate.rs
```diff
@@ -15,25 +15,38 @@
...
+	CError, ChildReference, nibble::LeftNibbleSlice, nibble_ops::NIBBLE_LENGTH, NibbleSlice,
+	node::{NodeHandle, NodeHandlePlan, NodePlan, OwnedNode, Value, ValuePlan}, NodeCodec, Recorder,
 	Result as TrieResult, Trie, TrieError, TrieHash,
-	TrieLayout,
+	TrieLayout, Record,
 };
 
+enum ValueMgmt {
+	/// The value should be omitted in the generated proof.
+	OmitValue,
```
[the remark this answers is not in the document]
cheme: This variant is an excluded value (when in the input iterator of known values) that is not a hashed node, AttachNode is the variant that is a hash node (I will rename it to 'OmitHashedValue').

trie-db/src/proof/generate.rs
```diff
@@ -15,25 +15,38 @@
...
+	/// Whether the value should be omitted and an additional node added
+	/// afterward. In this case the node is prefixed with `ESCAPE_HEADER`
+	/// to be disinguishible with `omit_value` case.
```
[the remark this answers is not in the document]
cheme: I need to remove it, the additional node is only in 'trie_codec.rs' (I did implement things wrongly at first and this comment was not updated).

trie-db/src/lib.rs
```diff
@@ -377,52 +403,68 @@ where
...
 	/// If true, the trie will allow empty values into `TrieDBMut`
 	const ALLOW_EMPTY: bool = false;
+
 	/// Hasher to use for this trie.
 	type Hash: Hasher;
 	/// Codec to use (needs to match hasher and nibble ops).
 	type Codec: NodeCodec<HashOut=<Self::Hash as Hasher>::Out>;
+
+	/// Alternate hashing threshold to apply on node change.
+	fn alt_threshold(&self) -> Option<u32>;
```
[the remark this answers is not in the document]
cheme: Sounds good, I will use `max_inline_value` if ok.

test-support/reference-trie/src/lib.rs
```diff
@@ -31,26 +32,64 @@ use trie_db::{
...
-use keccak_hasher::KeccakHasher;
 
 use trie_db::{
 	nibble_ops, NodeCodec,
-	Trie, TrieConfiguration,
+	Trie, TrieConfiguration, TrieDB, TrieDBMut,
 	TrieLayout, TrieMut,
 };
 pub use trie_root::TrieStream;
+use trie_root::Value as RValue;
```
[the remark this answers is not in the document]
cheme: Bad short for root of trie_root. (I did redefine the enum in trie_root, it differs a bit, but I mainly did it to keep trie_root a stand alone crate).
Will use `TrieStreamValue`.

trie-db/src/iter_build.rs
```diff
@@ -291,16 +313,29 @@ pub fn trie_visit<T, I, A, B, F>(input: I, callback: &mut F)
 			// one single element corner case
 			let (k2, v2) = previous_value;
 			let nkey = NibbleSlice::new_offset(&k2.as_ref()[..], last_depth);
-			let encoded = T::Codec::leaf_node(nkey.right(), &v2.as_ref()[..]);
 			let pr = NibbleSlice::new_offset(
 				&k2.as_ref()[..],
 				k2.as_ref().len() * nibble_ops::NIBBLE_PER_BYTE - nkey.len(),
 			);
+
+			let hashed;
```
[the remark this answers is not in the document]
cheme: I needed this because hash in value  is only borrowed.

arkpar (approved):

arkpar:
@bkchr Please review or assign someone

cheme:
@arkpar https://github.com/paritytech/trie/pull/142/commits/159e96469b7ec37549eceafa29fddaa8eb57ba11 is changing the prefix of detached value (hash db prefix to avoid collision).
Previously I was using the prefix of its node, but realize it would not work if the node containing the value does change (split partial or fuse with parent).
So this commit uses the full key as hash db prefix for detached value node.

trie-db/src/fatdb.rs
```diff
@@ -71,6 +73,14 @@ where
 	> {
 		FatDBIterator::<L>::new(&self.raw).map(|iter| Box::new(iter) as Box<_>)
 	}
+
+	fn key_iter<'a>(&'a self) -> Result<
+		Box<dyn TrieIterator<L, Item = TrieKeyItem<TrieHash<L>, CError<L>>> + 'a>,
+		TrieHash<L>,
+		CError<L>,
+	> {
+		unimplemented!()
```
thiolliere, line 82: why is it ok not to implement it ?

trie-db/src/fatdbmut.rs
```diff
@@ -80,27 +81,30 @@ where
...
 		// remove if it already exists.
-		if out.is_some() {
-			let aux_hash = L::Hash::hash(hash.as_ref());
-			self.raw.db_mut().remove(&aux_hash, EMPTY_PREFIX);
+		match &out {
+			Value::NoValue => (),
+			_ => {
+				let aux_hash = L::Hash::hash(hash.as_ref());
+				self.raw.db_mut().remove(&aux_hash, EMPTY_PREFIX);
+			},
```
thiolliere, line 107: sidenote: the condition could have been written `!matches(&out, Value::NoBalue)`

trie-db/src/fatdb.rs
cheme: Both fatdb and sectriedb (same for the mut variants) are specific to ethereum client.
It certainly could be implemented just by added the key iterator variant struct, but it is not very helpfull with ethereum (do not use a hashed value). So the only reason is to avoid a few line of code, I can add it or we can wait until someone needs it.

trie-db/src/lib.rs
```diff
@@ -377,52 +403,69 @@ where
...
 	const ALLOW_EMPTY: bool = false;
+
 	/// Hasher to use for this trie.
 	type Hash: Hasher;
 	/// Codec to use (needs to match hasher and nibble ops).
 	type Codec: NodeCodec<HashOut=<Self::Hash as Hasher>::Out>;
+
+	/// Threshold above which an external node should be
+	/// use to store a node value.
+	fn max_inline_value(&self) -> Option<u32>;
```
thiolliere, line 423: out of curiosity why not using a const here ?

trie-db/src/lib.rs
```diff
@@ -377,52 +403,69 @@ where
 /// Trait with definition of trie layout.
 /// Contains all associated trait needed for
 /// a trie definition or implementation.
-pub trait TrieLayout {
+/// Structure implementing this trait can define contextual behavior
+/// for trie, but should be small (clone cost should be insignifiant).
+pub trait TrieLayout: Default + Clone {
```
thiolliere, line 408: why do we bound `Default` here, isn't it error prone ?

I see for instance that `SecTrieDB::new` makes use of the default layout, I don't really know SecTrieDB but couldn't all those `new` which uses default layout be misused, and were actually meant to be called with a specific layout ?

Anyhow as far as I understand the code for now I feel like making the max_inline_value a const could solve most of this ? But maybe I miss the actual reason for all of this.

trie-db/src/lib.rs
cheme: It is mostly related to past design.
With current substrate usage, I can effectively choose a different layout when calling storage root, state-machine should allow that.
Then I just have to use two different layouts.
This would mean less change in substrate side (no need to pass around an instantiated layout boolean).
I think I will start a branch with the change and update if ok, with current code it will not fondamentally change things, but making a less permissive api may be good.

trie-db/src/lib.rs
cheme: Yes default in trie crate is old layout while default in substrate pr is new layout.

> Anyhow as far as I understand the code for now I feel like making the max_inline_value a const could solve most of this ? But maybe I miss the actual reason for all of this.

Reach the same conclusion (at first we got the layout behaviour from a state in the trie, so we could not have const), will change it today and come back to you if there is an unexpected issue.

trie-db/src/fatdb.rs
thiolliere: yes ok, in general I think that having half implemented trait is very error prone, even more when it is not documented, maybe we could have a new trait for this key iteration stuff, and only implement it where we need it.

trie-db/src/lib.rs
```diff
@@ -377,52 +403,69 @@ where
...
 /// used to allow switching implementation.
 pub trait TrieConfiguration: Sized + TrieLayout {
 	/// Operation to build a trie db from its ordered iterator over its key/values.
-	fn trie_build<DB, I, A, B>(db: &mut DB, input: I) -> <Self::Hash as Hasher>::Out where
-	DB: HashDB<Self::Hash, usize>,
-	I: IntoIterator<Item = (A, B)>,
-	A: AsRef<[u8]> + Ord,
-	B: AsRef<[u8]>,
+	fn trie_build<DB, I, A, B>(&self, db: &mut DB, input: I) -> <Self::Hash as Hasher>::Out where
+		DB: HashDB<Self::Hash, DBValue>,
```
thiolliere, line 442: why having changed the the HashDB value from usize to DBValue ?

trie-db/src/lookup.rs
```diff
@@ -118,7 +144,7 @@ where
 				// check if new node data is inline or hash.
 				match next_node {
 					NodeHandle::Hash(data) => {
-						hash = decode_hash::<L::Hash>(data)
+						self.hash = decode_hash::<L::Hash>(data)
```
thiolliere, line 147: why do we even assign it if we are going to break anyway ?
```suggestion
						decode_hash::<L::Hash>(data)
```

trie-db/src/lookup.rs
```diff
@@ -29,25 +29,49 @@ pub struct Lookup<'a, L: TrieLayout, Q: Query<L::Hash>> {
...
 		mut self,
 		key: NibbleSlice,
 	) -> Result<Option<Q::Item>, TrieHash<L>, CError<L>> {
 		let mut partial = key;
-		let mut hash = self.hash;
 		let mut key_nibbles = 0;
 
+		let mut full_key = key.clone();
+		full_key.advance(key.len());
+		let full_key = full_key.left();
```
thiolliere, line 70: I don't understand this code, isn't key exactly same as full_key ? why not just this:
```suggestion
		let mut full_key = key.clone();
```

cheme: key will be pointing to a different position in the slice than full_key.
Actually we advance by key base len so fullkey is the a nibble slice with position at end (size 0), and call to `left` will return the whole initial key.
It is the same as getting whole data used by the nibble slice, just there is no api for it.
(Value node are using their own full key as prefix so (b"key_to_value", None)).

thiolliere: ok, makes sense

trie-db/src/proof/generate.rs
```diff
@@ -210,7 +245,7 @@ impl<'a, C: NodeCodec> StackEntry<'a, C> {
 				);
 				&mut hash.as_mut()[..encoded_child.len()].copy_from_slice(encoded_child);
```
thiolliere, line 246: this would fix the warning
```suggestion
				hash.as_mut()[..encoded_child.len()].copy_from_slice(encoded_child);
```

trie-db/src/proof/generate.rs
```diff
@@ -75,13 +88,34 @@ impl<'a, C: NodeCodec> StackEntry<'a, C> {
 
 	/// Encode this entry to an encoded trie node with data properly omitted.
 	fn encode_node(mut self) -> TrieResult<Vec<u8>, C::HashOut, C::Error> {
+		let omit_value = self.omit_value;
+		let prepend_on_hashed = |mut encoded: Vec<u8>| -> Vec<u8> {
+			if let Some(header) = C::ESCAPE_HEADER {
+				if matches!(omit_value, ValueMgmt::OmitHashedValue) {
+					for i in header.iter() {
+						encoded.insert(0, *i);
+					}
```
thiolliere, line 97: is it intentional to write it in reversed ?

thiolliere: also inserting at index 0 isn't very efficient, maybe it is better to create a new vec rather than moving all elements by one every time.

trie-db/src/proof/generate.rs
```diff
@@ -75,13 +88,34 @@ impl<'a, C: NodeCodec> StackEntry<'a, C> {
...
+			}
+		};
+		let encoded = match self.node.node_plan() {
 			NodePlan::Empty => node_data.to_vec(),
-			NodePlan::Leaf { .. } if !self.omit_value => node_data.to_vec(),
+			NodePlan::Leaf { .. } if matches!(omit_value, ValueMgmt::Standard) => node_data.to_vec(),
 			NodePlan::Leaf { partial, value: _ } => {
 				let partial = partial.build(node_data);
-				C::leaf_node(partial.right(), &[])
+				prepend_on_hashed(C::leaf_node(partial.right(), Value::Value(&[])))
```
thiolliere, line 118: can a LeafNode contains a HashedValue ?

trie-db/src/node.rs
```diff
@@ -40,21 +41,52 @@ pub fn decode_hash<H: Hasher>(data: &[u8]) -> Option<H::Out> {
...
+
 /// Type of node in the trie and essential information thereof.
 #[derive(Eq, PartialEq, Clone)]
 #[cfg_attr(feature = "std", derive(Debug))]
 pub enum Node<'a> {
 	/// Null trie node; could be an empty root or an empty branch entry.
 	Empty,
 	/// Leaf node; has key slice and value. Value may not be empty.
-	Leaf(NibbleSlice<'a>, &'a [u8]),
+	Leaf(NibbleSlice<'a>, Value<'a>),
```
thiolliere, line 82: in the past the leaf node couldn't contain no value. Now it can, also can a leaf node contain a hashed value ?
It seems to me this should use another type LeafValue with only 1 or 2 variant (depending if it can contains hashed value which I don't know yet).

trie-db/src/node.rs
```diff
@@ -122,7 +178,7 @@ pub enum NodePlan {
 	/// Leaf node; has a partial key plan and value.
 	Leaf {
 		partial: NibbleSlicePlan,
-		value: Range<usize>,
+		value: ValuePlan,
```
thiolliere, line 181: same as node, previous type was more constrained than now.

trie-db/src/proof/generate.rs
cheme: It is adding a escape header sequence.
Compact proof got a special encoding for node that are skipped.
Value skipped are 0 length encoded, and build again when verifying (we verify with the value as parameter).
Problem is we do not know if the value is hashed or not (as long as the trie state can be hybrid), so when hashed
we prepend this sequence).
In practice this sequence is only one byte.

trie-db/src/proof/generate.rs
cheme: Yes, here we remove the value because it is a skippend one (in input when verifying), the matching above should be use when we do not skip it.

trie-db/src/proof/generate.rs
cheme: I am just thinking, the fact that we will migrate state, we can maybe remove this code and apply threshold when verifying to know if we want to hash?

thiolliere: > In practice this sequence is only one byte.

I now read that in verify we check the header is in order, I think we should just make ESCAPE_HEADER a `Option<u8>` to fix this

trie-db/src/node.rs
cheme: Leaf can contain hashed value (it is as needed as for the branch case).
But by definition a Leaf cannot contain no value (otherwhise it would be removed).

Most of the code for Leaf without value leads to panic (unreachable probably, I did not really check), or at least it cannot encode or decode.
I did not want to have two different Value enum (we got other enums for Value when data is owned or for decode plan).
But now that the code is a bit more reviewed, maybe using `LeafValue` could make sense, would just make more code.

trie-db/src/proof/generate.rs
cheme: Oh yes, you are right insert should be rev, I will switch to single byte (if needed sequence could be use in the future).

thiolliere: > I am just thinking, the fact that we will migrate state, we can maybe remove this code and apply threshold when verifying to know if we want to hash?

But until it is all migrated we still need this no ?

cheme: This proof is not use a lot, the only part of code I know that use it is fastsync or warpsynch, but some of their use case do not really work with hybrid.
I don't know of any other usage of this compact proof (the one for cumulus is trie_codec.rs version which really need escaping).

trie-db/src/proof/generate.rs
```diff
@@ -480,28 +574,14 @@ fn match_key_to_branch_node<'a, 'b, C: NodeCodec>(
...
 /// encoded into proof nodes and added to the finalized proof.
-fn unwind_stack<C: NodeCodec>(
-	stack: &mut Vec<StackEntry<C>>,
+fn unwind_stack<L: TrieLayout>(
+	stack: &mut Vec<StackEntry<L::Codec>>,
 	proof_nodes: &mut Vec<Vec<u8>>,
 	maybe_key: Option<&LeftNibbleSlice>,
-) -> TrieResult<(), C::HashOut, C::Error>
-{
+) -> TrieResult<(), TrieHash<L>, CError<L>> {
```
thiolliere, line 584: those changes look not necessary, is it better usage of types than before ?

trie-db/src/node.rs
thiolliere: we could also keep the old way and have Value here, and `Option<Value>` for branches maybe. and remove the variant NoValue

cheme:
https://github.com/paritytech/trie/pull/142/commits/910c79222905b9da18303c6382dea0e8c7e16941 does revert to former api without trie layout instantiated, since it is not needed anymore (just finished checking on substrate branch).

trie-db/src/lib.rs
```diff
@@ -384,44 +403,58 @@ pub trait TrieLayout {
...
+/// Small enum indicating representation of a given children.
+pub enum ChildrenDecoded {
+	/// Children is stored as a node.
+	Hash,
+	/// Children is stored as an inline node.
+	Inline,
+	/// No children is present.
+	None,
+}
+
```
thiolliere, line 425: Where is this struct used ?

```suggestion
```

trie-db/src/fatdb.rs
thiolliere: looking again, I think we should just implement it, even in a non efficient way is fine. by using iter underneath.

trie-db/src/triedb.rs
```diff
@@ -142,6 +155,14 @@ where
 	> {
 		TrieDBIterator::new(self).map(|iter| Box::new(iter) as Box<_>)
 	}
+
+	fn key_iter<'a>(&'a self)-> Result<
```
thiolliere, line 159: sidenote: this is not needed for the PR isn't it ? maybe the optimization could be moved to another PR, that would make review easier to me at least

thiolliere: thinking again, maybe it is useful for big keys

cheme: Main purpose of the specific key iterartor is to not access values, so we don't have values in proofs but only their hashes when unaccessed.
(that is what next_storage_key need to use in substrate)

trie-db/src/triedb.rs
```diff
@@ -317,24 +414,80 @@ impl<'a, L: TrieLayout> Iterator for TrieDBIterator<'a, L> {
...
+								let mut res = TrieHash::<L>::default();
+								res.as_mut().copy_from_slice(hash);
+								return Some(Err(Box::new(
+									TrieError::IncompleteDatabase(res)
+								)));
+							}
+						},
+						Value::Value(value) => value.to_vec(),
+						Value::HashedValue(_hash, Some(value)) => value,
+						Value::NoValue => return None,
```
thiolliere:
```suggestion
						 // Should never happen as it has continued above
						Value::NoValue => return None,
```

cheme:
@thiolliere , https://github.com/paritytech/trie/pull/142/commits/24c9ae9d5555f5dd80dcc02d8c4a751354a9fd49 switch to using Option<Value> instead of Value.
It is better for some aspect, not for others, I got no strong opinion about this change, so will probably keep it.

trie-db/src/lib.rs
cheme: :+1: Probably some uncleaned code, removed it.

trie-db/src/node.rs
thiolliere: I find it better now

cheme:
https://github.com/paritytech/trie/pull/142/commits/b7238ef0a5550e22b3e061b72d714c6cffafc5d1 does remove the hybrid specific compact proof encoding.
@thiolliere , ESCAPE_HEADER was not remove, it is still needed in trie_codec.rs (when header is present we got attached value after the hash). I cannot remove it because the codec is not writing value length when encoding a hashed value so we need to know if value is in proof or not.

trie-db/src/proof/generate.rs
cheme: I did revert them.

trie-db/src/lib.rs
cheme: Was not really working with usize, except for tests. I think this function is used nowhere, could just be removed.

trie-db/src/fatdb.rs
cheme: My main issue was to add unneeded code to the codebase, but I did implement it for sectrie, so https://github.com/paritytech/trie/pull/142/commits/99e83d3272879790c44bce0b2619b1ab4cc642f3 implement that.

trie-root/src/lib.rs
```diff
@@ -51,16 +79,16 @@ pub trait TrieStream {
...
 	);
 	/// Append an empty child node. Optional.
 	fn append_empty_child(&mut self) {}
 	/// Wrap up a Branch node portion of a `TrieStream` and append the value
 	/// stored on the Branch (if any).
-	fn end_branch(&mut self, _value: Option<&[u8]>) {}
+	fn end_branch(&mut self, _value: Value) {}
 	/// Append a Leaf node
-	fn append_leaf(&mut self, key: &[u8], value: &[u8]);
+	fn append_leaf(&mut self, key: &[u8], value: Value);
```
thiolliere, line 91: I would still prefer to have a type without NoValue here, that would prevent having to write unreachable in the TrieStream implementation in substrate.

cheme: :+1: I did forgot about trie_root, will switch it too

trie-db/src/triedbmut.rs
```diff
@@ -1444,17 +1578,27 @@ where
...
+						if let Some(value) = value_node {
+							let value_hash = self.db.insert(k.as_prefix(), value);
+							k.drop_lasts(mov);
+							return ChildReference::Hash(value_hash);
+						}
 						let cr = self.commit_child(child, &mut k);
 						k.drop_lasts(mov);
 						cr
-					}
+					},
```
thiolliere, line 1593: this callback function looks quite weird, either value_node is defined and it is supposed to create the childreference of the "value_node" or value_node is none and it is supposed to create the childreference for a regular node.

When calling this in the implementation Value::into_encoded we have to give a fake child parameter if I understand correclty.

IMHO we should create a new type for first argument:
```
enum ChildCreation {
ValueNode(value: &[u8]),
RegularNode(TrieHandle),
}
```
This way I think usage would make more sense

trie-db/src/triedbmut.rs
```diff
@@ -66,23 +67,119 @@ fn empty_children<H>() -> Box<[Option<NodeHandle<H>>; 16]> {
...
+			Value::NewHashedValue(None, _value) => unreachable!("New external value are always added before encoding anode"),
+		};
+		value
+	}
+
+	fn inline_fetched_value(&self) -> DBValue {
+		match self {
+			Value::Value(value) => value.clone(),
+			Value::NewHashedValue(_, value) => value.clone(),
+			Value::HashedValue(_, Some(value)) => value.clone(),
```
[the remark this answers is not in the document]
cheme: Yes unreachable too, just the third one was problematic to implement.

trie-db/src/triedbmut.rs
```diff
@@ -66,23 +67,124 @@ fn empty_children<H>() -> Box<[Option<NodeHandle<H>>; 16]> {
...
+		f: &mut F,
+	) -> EncodedValue<'a>
+	where
+		F: FnMut(NodeToEncode<TrieHash<L>>, Option<&NibbleSlice>, Option<u8>) -> ChildReference<TrieHash<L>>,
+	{
+		if let Value::NewHashedValue(hash, value) = self {
+			let new_hash = if let ChildReference::Hash(hash) = f(NodeToEncode::ValueNode(value.as_slice()), partial, None) {
+				hash
+			} else {
+				unreachable!("Passing a hash as parameter is only to add an attached value")
```
thiolliere, line 144:
```suggestion
				unreachable!("Value node can never be inlined; qed")
```

trie-db/src/triedbmut.rs
```diff
@@ -873,7 +1025,7 @@ where
 					// make an extension using it. this is a replacement.
 					InsertAction::Replace(Node::Extension(
 						existing_key.to_stored_range(common),
-						self.storage.alloc(Stored::New(augmented_low)).into()
+						self.storage.alloc(Stored::New(augmented_low)).into(),
 					))
 				}
 			},
```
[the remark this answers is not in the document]
cheme: 'fix' is checking a node is well formed, so I expect it can.
Removing the value of a branch is potentially making your branch 'one child no value' and replacing it by a leaf node. In this case if we run with eth variant, the extension before the node will need to be remove. I am not sure if the actual change is in fix or in fix sometime and in the remove visitor some other time.
But the extension case should only be touched when `TrieLayout` trait has associated constant `USE_EXTENSION` to true (which only happen with parity ethereum layout).

trie-db/src/proof/generate.rs
```diff
@@ -75,13 +88,24 @@ impl<'a, C: NodeCodec> StackEntry<'a, C> {
...
+			if matches!(omit_value, ValueMgmt::Standard) {
+				Some(value_range.build(&node_data))
+			} else {
+				None
+			}
+		};
+		let encoded = match self.node.node_plan() {
 			NodePlan::Empty => node_data.to_vec(),
-			NodePlan::Leaf { .. } if !self.omit_value => node_data.to_vec(),
+			NodePlan::Leaf { .. } if matches!(omit_value, ValueMgmt::Standard) => node_data.to_vec(),
```
thiolliere, line 105: I'm not sure I see the point for the 3 variants enum, because in this code we only match against Standard, both here and just above line 97.
why not just keeping a bool here ?

trie-db/src/node.rs
```diff
@@ -40,21 +41,45 @@ pub fn decode_hash<H: Hasher>(data: &[u8]) -> Option<H::Out> {
...
+#[cfg_attr(feature = "std", derive(Debug))]
+pub enum Value<'a> {
+	/// Value byte slice.
+	Value(&'a [u8]),
+	/// Hash byte slice and the actual value if accessed.
+	HashedValue(&'a [u8], Option<DBValue>),
+}
+
+impl<'a> Value<'a> {
+	pub(crate) fn new(value: &'a [u8], threshold: Option<u32>) -> Option<Self> {
```
thiolliere, line 55: maybe we could improve the name here, like Value::Value could be renamed Value::Inlined, and Value::HashedValue could be renamed Value::ValueNode,

and `new` which only tries to create a `Value::Value` could be renamed `new_inlined`.

That said I'm not sure of the name I proposed either. so maybe better to keep as it is

PR #144, opened 2021-11-05
Add rustfmt.toml + run rustfmt

bkchr:


cheme (approved):
I don't always understand some of rustfmt behavior.
But the crate had a lot of existing formatting inconsistency, so I will be happy happy to do the switch.

PR #146, opened 2021-12-09
Adding support for eip-1186 proofs

lightyear15:
EIP-1186[1] is the standard for generating proofs used by EVM-based clients.

The current proof crate in trie-db generates (and verifies) compact proofs that are not compatible with eip-1186.
This commit adds functions to generate and verify such proofs, adding also proper testing of such new APIs

[1] https://eips.ethereum.org/EIPS/eip-1186

### DIFF
--- a/trie-db/CHANGELOG.md
+++ b/trie-db/CHANGELOG.md
@@ -5,7 +5,6 @@ The format is based on [Keep a Changelog].
 [Keep a Changelog]: http://keepachangelog.com/en/1.0.0/
 
 ## [Unreleased]
-Support eip 1186 trie proofs. [#146](https://github.com/paritytech/trie/pull/146)
 
 ## [0.23.0] - 2021-10-19
 - Support for value nodes. [#142](https://github.com/paritytech/trie/pull/142)
--- a/trie-db/CHANGELOG.md
+++ b/trie-db/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog].
 [Keep a Changelog]: http://keepachangelog.com/en/1.0.0/
 
 ## [Unreleased]
+Support eip 1186 trie proofs. [#146](https://github.com/paritytech/trie/pull/146)
 
 ## [0.23.0] - 2021-10-19
 - Support for value nodes. [#142](https://github.com/paritytech/trie/pull/142)
--- a/trie-db/src/proof/eip1186.rs
+++ b/trie-db/src/proof/eip1186.rs
@@ -1,0 +1,305 @@
+use crate::{
+	nibble::NibbleSlice,
+	node::{Node, NodeHandle, Value, decode_hash},
+	recorder::Recorder,
+	rstd::{result::Result, vec::Vec},
+	CError, NodeCodec, Result as TrieResult, Trie, TrieHash, TrieLayout,
+};
+use hash_db::Hasher;
+
+/// Generate an eip-1186 compatible proof for key-value pairs in a trie given a key.
+pub fn generate_proof<T, L>(
+	trie: &T,
+	key: &[u8],
+	) -> TrieResult<(Vec<Vec<u8>>, Option<Vec<u8>>), TrieHash<L>, CError<L>>
+where
+T: Trie<L>,
+L: TrieLayout,
+{
+	let mut recorder = Recorder::new();
+	let item = trie.get_with(key, &mut recorder)?;
+	let proof: Vec<Vec<u8>> = recorder.drain().into_iter().map(|r| r.data).collect();
+	Ok((proof, item))
+}
+
+/// Errors that may occur during proof verification. Most of the errors types simply indicate that
+/// the proof is invalid with respect to the statement being verified, and the exact error type can
+/// be used for debugging.
+#[derive(PartialEq, Eq)]
+#[cfg_attr(feature = "std", derive(Debug))]
+pub enum VerifyError<'a, HO, CE> {
+	/// The proof does not contain any value for the given key
+	/// the error carries the nibbles left after traversing the trie
+	NonExistingValue(NibbleSlice<'a>),
+	/// The proof contains a value for the given key
+	/// while we were expecting to find a non-existence proof
+	ExistingValue(Vec<u8>),
+	/// The proof indicates that the trie contains a different value.
+	/// the error carries the value contained in the trie
+	ValueMismatch(Vec<u8>),
+	/// The proof is missing trie nodes required to verify.
+	IncompleteProof,
+	/// The node hash computed from the proof is not matching.
+	HashMismatch(HO),
+	/// One of the proof nodes could not be decoded.
+	DecodeError(CE),
+	/// Error in converting a plain hash into a HO
+	HashDecodeError(&'a[u8])
+}
+
+#[cfg(feature = "std")]
+impl<'a, HO: std::fmt::Debug, CE: std::error::Error> std::fmt::Display for VerifyError<'a, HO, CE> {
+	fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+		match self {
+			VerifyError::NonExistingValue(key) => write!(f, "Key does not exist in trie: reaming key={:?}", key),
+			VerifyError::ExistingValue(value) => write!(f, "trie contains a value for given key value={:?}", value),
+			VerifyError::ValueMismatch(key) => write!(f, "Expected value was not found in the trie: key={:?}", key),
+			VerifyError::IncompleteProof => write!(f, "Proof is incomplete -- expected more nodes"),
+			VerifyError::HashMismatch(hash) => write!(f,"hash mismatch found: hash={:?}",hash),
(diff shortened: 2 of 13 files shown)


trie-db/src/proof/eip1186.rs
```diff
@@ -0,0 +1,305 @@
...
+				Err(VerifyError::IncompleteProof)
+			} else {
+				key.advance(1);
+				let new_root = decode_hash::<L::Hash>(hash).ok_or(VerifyError::HashDecodeError(hash))?;
+				process_node::<L>(&new_root, &proof[0], key, expected_value, &proof[1..])
+			}
+		}
+		Some(Some(NodeHandle::Inline(encoded_node))) => {
+			key.advance(1);
+			let root = L::Hash::hash(encoded_node);
```
cheme, line 253: This hash calculation is not needed.

trie-db/src/proof/eip1186.rs
```diff
@@ -0,0 +1,305 @@
...
+{
+	if proof.is_empty() {
+		return Err(VerifyError::IncompleteProof);
+	}
+	let key = NibbleSlice::new(raw_key);
+	process_node::<L>(root, &proof[0], key, expected_value, &proof[1..])
+}
+
+fn process_node<'a, L>(
+	root: &<L::Hash as Hasher>::Out,
```
cheme, line 95: I would call it `expected_node_hash` or something that relates to a node (root makes me think it is the root node hash).

trie-db/test/src/eip1186.rs
```diff
@@ -0,0 +1,184 @@
...
+		}
+		(db, root)
+	};
+	// Generate proof for the given keys..
+	let trie = <TrieDB<L>>::new(&db, &root).unwrap();
+	let proof = generate_proof::<_, L>(&trie, key).unwrap();
+	(root, proof.0, proof.1)
+}
+
+test_layouts!(trie_proof_works2, trie_proof_works_internal2);
```
cheme, line 69: :+1: for running on all layout.

cheme (commented):
Code looks good to me. Nice to see it. 
Using its own module, it is not obtrusive and could probably be in his own crate (may need to expose some internal structs or methods).

Next step could be to implement it to run over multiple keys.

But I am just not sure if it should be include in the code base.
On one hand I am not sure about the state of this eip (@arkpar  or @sorpaas maybe you did know).

On the other hand I am wondering if the additional code is really required: in parity-ethereum/open-ethereum we just did load all encoded nodes in a memory-db (which does the hash calculation and is queried by this hash) and run the key query on a `TrieDB` that uses this partial `MemoryDB`.
For me, the additional value of this PR is that it enforces nodes to be ordered by key on verification and it provides more error information (but I do not know if eip requires it).
Also it is more efficient than using a `MemoryDB` because we do not need to move/copy the encoded nodes to the 'memory-db' map.

If those point are really needed (ordering check, errors and performance), modifying https://github.com/paritytech/trie/blob/master/trie-db/src/proof/verify.rs to keep the useless hashes may also be a good way to factor code a bit and already have `multiple key` case managed. But as I mentioned I am not sure this is needed (using a `TrieDB` over a partial `MemoryDB` build from proof nodes is really what I prefer as it is more flexible).

trie-db/src/proof/eip1186.rs
```diff
@@ -0,0 +1,305 @@
...
+		(None, _, Some(_)) => Err(VerifyError::NonExistingValue(key)),
+		(Some(Value::Inline(inline_data)), _, Some(value)) => {
+			if inline_data == value {
+				Ok(())
+			} else {
+				Err(VerifyError::ValueMismatch(inline_data.to_vec()))
+			}
+		},
+		(Some(Value::Inline(inline_data)), _, None) => Err(VerifyError::ExistingValue(inline_data.to_vec())),
+		(Some(Value::Node(plain_hash, _)), Some(next_proof_item), Some(value)) => {
```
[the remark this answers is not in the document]
lightyear15: Is this comment referring line 287 or 286?
If we leave the case as it is, even if eth does not use such node type, is it going to cause any issue in proof verification?
Can we leave it as it is?

trie-db/src/proof/eip1186.rs
lightyear15: Sorry,
I am not sure I understand your comment. ``root`` is passed to ``process_node`` function the line after.
How can I avoid this hashing operation?

trie-db/src/proof/eip1186.rs
lightyear15: :+1:  I will also change ``calculate_root`` to ``calculated_node_hash`` then.

trie-db/src/proof/eip1186.rs
cheme: IIReadC, in `process_node` this calculated hash is only compared again with encoded hash.
Making the input in `process_node` optional (or with an `enum { Inline, Node(H) }`) would work to avoid this check for inline node.

trie-db/src/proof/eip1186.rs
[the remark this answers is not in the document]
cheme: Sorry just refering to handling the `Value::Node` case.
`Value::Node` is never use with ethereum trie layout, values are always inlined in either a branch or a leaf (also there is no NibbledBranch with ethereum codec).
But the code you write looks good to me (adding the value node as a following node is what I did for the other proofs), so I'd rather keep it (it allows running the tests on all layout which looks nice).

lightyear15:
Hello @cheme ,
is there any more action required from me to have this PR merged?

arkpar:
I'm fine with merging it here. I think this crate should be forked for use in substrate eventually. The trie formats between ethereum and substrate have diverged significantly and will probably diverge further. Maintaining a single codebase requires too many abstractions already.

cheme:
About current PR code there is a few point that should be done:
- remove warnings (at least no warnings when running cargo test).
- run `cargo +nightly fmt --all` (need cargo format and nightly toolchain, I can also run it).
- not hashing inline node would be more correct (we never should hash an inline node).

About:
> modifying https://github.com/paritytech/trie/blob/master/trie-db/src/proof/verify.rs to keep the useless hashes may also be a good way to factor code a bit and already have multiple key case managed.

I just checked this, it is totally doable (same ordering of nodes), basically create an optional mode where instead of filling nodes with omitted elements (value or children hash), the elements are checked (hash of children and values).
Could be a way to reduce the code, but I also realize it is not very straightforward,  so probably not a good idea from my part.

lightyear15:
@arkpar @cheme ,
thanks a lot for your feedbacks.
@cheme to your points:
- I don't see any warnings where running cargo test, let me know if I am overlooking something please.
- I hope I will get cargo fmt ok at some point, easier said than done
- I changed the ``process_node`` function to accept an optional root hash, so whenever the hash is not available I can pass a a ``None`` rather than doing pointless hash computation.

cheme:
> I don't see any warnings where running cargo test, let me know if I am overlooking something please.

second round fixes the unused imports, there is still one, but unrelated to your PR (probably new rustc compiler finding new warning).

I forgot this,  can you include:
```diff
diff --git a/trie-db/CHANGELOG.md b/trie-db/CHANGELOG.md
index 98aa56e..739ba7f 100644
--- a/trie-db/CHANGELOG.md
+++ b/trie-db/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog].
 [Keep a Changelog]: http://keepachangelog.com/en/1.0.0/
 
 ## [Unreleased]
+- Support eip 1186 trie proofs. [#146](https://github.com/paritytech/trie/pull/146)
 
 ## [0.23.0] - 2021-10-19
 - Support for value nodes. [#142](https://github.com/paritytech/trie/pull/142)
```

> I like squashing away minor commits before merging...

Don't worry, I will use github `squash and merge`, I will probably just write `Support eip 1186 trie proofs`.

sorpaas (changes requested):
Can you move the code to a separate crate (maybe call it `trie-eip1186`) or is there a reason that you insist on the current code structure? Many current users will not need the 1186 feature and it saves compile time for them, while at the same time has no downside for 1186 users.

lightyear15:
> Can you move the code to a separate crate (maybe call it `trie-eip1186`) or is there a reason that you insist on the current code structure? Many current users will not need the 1186 feature and it saves compile time for them, while at the same time has no downside for 1186 users.

No particular reason, I thought it could have been a good follow-up PR, but if you think it's fundamental I can do it in this one.

trie-db/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog].
 [Keep a Changelog]: http://keepachangelog.com/en/1.0.0/
 
 ## [Unreleased]
+Support eip 1186 trie proofs. [#146](https://github.com/paritytech/trie/pull/146)
```
cheme, line 8:
```suggestion
```

sorpaas (approved):

lightyear15:
Is this PR waiting for something to happen on my side?

PR #153, opened 2022-02-21
Fix hex trace output

ggwpez:
The trie trace logs use one line per byte, making it unreadable.  
Instead put it in one line. 

There is no test for this but I tested it in the [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b19c89e680851fdbdf683b605d2b8227).

### DIFF
--- a/trie-db/src/triedbmut.rs
+++ b/trie-db/src/triedbmut.rs
@@ -1635,7 +1635,7 @@ where
 					}
 				});
 				#[cfg(feature = "std")]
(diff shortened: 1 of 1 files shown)


ggwpez:
It is still not really good readable, should I change it to `hex::encode(…)`?

cheme:
> It is still not really good readable, should I change it to `hex::encode(…)`?

Sounds good, maybe more `ToHex` as it is already use in the crate.

cheme (approved):

arkpar (approved):

PR #155, opened 2022-04-08
Do not check for root in `TrieDB` and `TrieDBMut` constructors

bkchr:
This leads to one extra storage lookup before we are doing the actual key lookup. This leads to reading the `root` twice
when just wanting to lookup one key in the trie. We also return an error on lookup anyway if `root` doesn't exists, so
there is really no need to do this twice.

### DIFF
--- a/trie-db/src/lib.rs
+++ b/trie-db/src/lib.rs
@@ -304,9 +304,8 @@ impl Default for TrieSpec {
 
 /// Trie factory.
 #[derive(Default, Clone)]
-pub struct TrieFactory<L: TrieLayout> {
+pub struct TrieFactory {
 	spec: TrieSpec,
-	layout: L,
 }
 
 /// All different kinds of tries.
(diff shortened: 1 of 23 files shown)


cheme:
I like the idea of not checking root (a trie is valid with any root and should be buildable, then missing is something else).

Probably the initial design was to fail early when state was not available: in system where the presence of root indicate that the whole state for the trie exists (I am thinking maybe previous ethereum code). 

Making the function a combination of 'new' and '~check_existing_state'.

Could failing early in substrate be good: we query one node and avoid all processing between triedb instantiation and actual first triedb access. But if the state is correct, we should not have this error ever.
Actually maybe when state is pruned, but then we don't try to sync the block so sounds ok.

Maybe we need to keep the old instantiation for compatibility (changing name and indicating it in the https://github.com/paritytech/trie/blob/master/trie-db/CHANGELOG.md), or add a small 'exists_root' function so previous.
Actually the Changelog could just say: old behavior through 'new' then 'db.contains(..)' check.

cheme:
Also I did find it inconvenient a few time to have to handle this error, so probably I would suggest adding a good CHANGELOG and this will be good for me.
(also ci failure for bench: two unwrap two remove and rust format needed but that is a detail)

trie-db/src/lib.rs
```diff
@@ -376,30 +375,27 @@ impl<'db, L: TrieLayout> Trie<L> for TrieKinds<'db, L> {
 	}
 }
 
-impl<'db, L> TrieFactory<L>
-where
-	L: TrieLayout + 'db,
-{
+impl TrieFactory {
```
cheme, line 378: `TrieFactory` is not used by substrate, but I feel like the change is good to have. Worth mentioning in CHANGELOG though.

trie-db/src/triedb.rs
```diff
@@ -62,22 +62,11 @@ impl<'db, L> TrieDB<'db, L>
...
-		if !db.contains(root, EMPTY_PREFIX) {
-			Err(Box::new(TrieError::InvalidStateRoot(*root)))
-		} else {
-			Ok(TrieDB { db, root, hash_count: 0 })
-		}
-	}
-
-	/// `new_with_layout`, but do not check root presence, if missing
-	/// this will fail at first node access.
-	pub fn new_unchecked(db: &'db dyn HashDBRef<L::Hash, DBValue>, root: &'db TrieHash<L>) -> Self {
```
cheme, line 80: Oh we actually have the unchecked version already. I never use it so new being unchecked by default makes sense.
Changelog could simply be `new_unchecked` replace `new`.

trie-db/src/lib.rs
bkchr: Here I just removed the warning of the unused `layout`

bkchr:
> Actually the Changelog could just say: old behavior through 'new' then 'db.contains(..)' check.

Yeah that sounds reasonable

trie-db/src/lib.rs
cheme: I mean api did change (better `new`).

cheme (approved):

arkpar (approved):

PR #162, opened 2022-07-30
fix fuzzer

cheme:
Just updating fuzzer to latest changes.


cheme:
This was added to some other PR

PR #181, opened 2023-01-30
Iterator refactoring

koute:
This PR simplifies and refactors the iterators in `trie-db`.

The changes can be summarized as follows:

* Added `TrieDBNodeIterator`-like `TrieDBRawIterator` and made it lifetimeless (instead of storing a `&TrieDB` it now takes it as a parameter during the iteration)
* `SuspendedTrieDBKeyIterator` was removed since the `TrieDBRawIterator` is a complete replacement for it (with the added bonus that it can be iterated on)
* `TrieDBIterator::next` and `TrieDBKeyIterator::next` were renamed and moved into `TrieDBRawIterator` with no changes to the logic (except making them use `&TrieDB` passed as a parameter instead of the one stored in `self`, and cleaning them up slightly)
* `TrieDBIterator` and `TrieDBKeyIterator` now use `TrieDBRawIterator` under the hood
* A bunch of `#[inline]`s were added to make recreating a `&TrieDB` on each iteration into a zero-cost operation
* All of the relevant crate versions were bumped

This makes further refactoring of storage iterators in `substrate` possible.

### DIFF
--- a/trie-db/src/iterator.rs
+++ b/trie-db/src/iterator.rs
@@ -331,7 +331,6 @@ impl<L: TrieLayout> TrieDBRawIterator<L> {
 		Result<(&NibbleVec, Option<&TrieHash<L>>, &Rc<OwnedNode<DBValue>>), TrieHash<L>, CError<L>>,
 	> {
 		enum IterStep<O, E> {
-			PopTrail,
 			Continue,
 			Descend(Result<(OwnedNode<DBValue>, Option<O>), O, E>),
 		}
@@ -363,7 +362,13 @@ impl<L: TrieLayout> TrieDBRawIterator<L> {
 								self.key_nibbles.drop_lasts(partial.len() + 1);
 							},
 						}
-						IterStep::PopTrail
+						self.trail.pop().expect(
+							"method would have exited at top of previous block if trial were empty;\
+								trial could not have been modified within the block since it was immutably borrowed;\
+								qed",
+						);
+						self.trail.last_mut()?.increment();
+						continue
 					},
 					(Status::At, NodePlan::Extension { partial: partial_plan, child }) => {
 						let partial = partial_plan.build(node_data);
@@ -408,14 +413,6 @@ impl<L: TrieLayout> TrieDBRawIterator<L> {
 			};
 
 			match iter_step {
-				IterStep::PopTrail => {
-					self.trail.pop().expect(
-						"method would have exited at top of previous block if trial were empty;\
-							trial could not have been modified within the block since it was immutably borrowed;\
-							qed",
-					);
-					self.trail.last_mut()?.increment();
-				},
 				IterStep::Descend::<TrieHash<L>, CError<L>>(Ok((node, node_hash))) => {
 					self.descend(node, node_hash);
 				},
--- a/trie-db/src/iterator.rs
+++ b/trie-db/src/iterator.rs
@@ -422,49 +422,47 @@ impl<L: TrieLayout> TrieDBRawIterator<L> {
 	///
 	/// Must be called with the same `db` as when the iterator was created.
 	pub fn next_item(&mut self, db: &TrieDB<L>) -> Option<TrieItem<TrieHash<L>, CError<L>>> {
-		while let Some(item) = self.next_raw_item(db) {
-			match item {
-				Ok((prefix, _, node)) => {
-					let mut prefix = prefix.clone();
-					let value = match node.node() {
-						Node::Leaf(partial, value) => {
-							prefix.append_partial(partial.right());
-							value
-						},
-						Node::Branch(_, value) => match value {
-							Some(value) => value,
-							None => continue,
-						},
-						Node::NibbledBranch(partial, _, value) => {
-							prefix.append_partial(partial.right());
-							match value {
-								Some(value) => value,
-								None => continue,
-							}
-						},
-						_ => continue,
-					};
-
-					let (key_slice, maybe_extra_nibble) = prefix.as_prefix();
-					let key = key_slice.to_vec();
-					if let Some(extra_nibble) = maybe_extra_nibble {
-						return Some(Err(Box::new(TrieError::ValueAtIncompleteKey(
-							key,
-							extra_nibble,
-						))))
-					}
-					let value = match value {
-						Value::Node(hash) =>
-							match Self::fetch_value(db, &hash, (key_slice, None)) {
-								Ok(value) => value,
-								Err(err) => return Some(Err(err)),
-							},
-						Value::Inline(value) => value.to_vec(),
-					};
-					return Some(Ok((key, value)))
-				},
+		while let Some(raw_item) = self.next_raw_item(db) {
+			let (prefix, _, node) = match raw_item {
+				Ok(raw_item) => raw_item,
 				Err(err) => return Some(Err(err)),
+			};
+
+			let mut prefix = prefix.clone();
+			let value = match node.node() {
+				Node::Leaf(partial, value) => {
+					prefix.append_partial(partial.right());
+					value
+				},
+				Node::Branch(_, value) => match value {
+					Some(value) => value,
+					None => continue,
+				},
+				Node::NibbledBranch(partial, _, value) => {
+					prefix.append_partial(partial.right());
+					match value {
+						Some(value) => value,
+						None => continue,
+					}
+				},
+				_ => continue,
+			};
+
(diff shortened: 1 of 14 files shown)


trie-db/src/triedb.rs
```diff
@@ -451,42 +439,39 @@ impl<'a, 'cache, L: TrieLayout> TrieDBIterator<'a, 'cache, L> {
...
+	pub fn into_raw(self) -> TrieDBRawIterator<L> {
+		self.raw_iter
 	}
 }
 
 impl<'a, 'cache, L: TrieLayout> TrieIterator<L> for TrieDBIterator<'a, 'cache, L> {
 	/// Position the iterator on the first element with key >= `key`
 	fn seek(&mut self, key: &[u8]) -> Result<(), TrieHash<L>, CError<L>> {
-		TrieIterator::seek(&mut self.inner, key)
+		self.raw_iter.seek_prefix(self.db, key).map(|_| ())
```
cheme, line 459: I am wondering if the function for raw_iter `seek_prefix` should be rename (I am maybe the one that give this ~ naming in the first place ). Here it just does a seek key which is what we need, but I had to check the code to be sure it is not the variant that seek a key and then remove nodes from the stack so it limits remaining iteration to the prefix.

cheme (approved):
About the PR, removing `&db` from Raw/Node iterator looks like a good change to me (no need for the suspended, and surely calling it raw makes it clear one can do something wrong with this iterator).

What are the plans for the substrate optim, I remember planning to use the suspended struct in host function to use it again if the start key used in a following next_key call is the same as the previously return key (but don't find back my branch)?

Or is there an idea to use a host api that keep state of iterator (I got in https://github.com/cheme/substrate/tree/transient a host api that keeps a hasher state in memory could be runing similarily)?

Regarding possible substrate optimization, there is one I would find interesting, that would be to switch sp-state-machine `backend` trait to use `&mut ` on all accesses, thus removing the need for inner mutability (for trie cache and recorder). I started in this direction in https://github.com/paritytech/substrate/pull/13006, but only covered `Externalities` trait (switching trie backend requires other changes, at least one existing clone to fetch runtime, but it can certainly be avoided). But I think this would not touch the trie crate (recorder and cache being already &mut).

cheme:
> TrieDBIterator and TrieDBKeyIterator now use TrieDBRawIterator under the hood

wasn't it the case already?

cheme:
> This makes further refactoring of storage iterators in substrate possible.

Does the suspended approach not work (I remember putting something in place but probably drop it at some point, not sure why)?

koute:
> What are the plans for the substrate optim, I remember planning to use the suspended struct in host function to use it again if the start key used in a following next_key call is the same as the previously return key (but don't find back my branch)?
> [...]
> Does the suspended approach not work (I remember putting something in place but probably drop it at some point, not sure why)?

For now I'm mostly just simplifying the way we do iteration; e.g. the `Backend` trait in `state-machine` has something like ~8 methods which do iteration, some using an `Fn` callback and some just collecting into a `Vec`, and I'm essentially replacing them all with a single method that just returns a raw iterator (plus two extra methods which return dedicated iterators for keys and pairs which are themselves just tiny wrappers around the raw iterator).

So it is essentially the suspend approach, it just needs these changes to make it fully possible (not everything was originally exported the first time I tried it IIRC) and simpler.

In general for now my main objective is to just simplify and remove dangerous methods (those which return a `Vec`, which make our RPC nodes explode), but I also want to eventually look into speeding up storage iteration/accesses altogether.

> Regarding possible substrate optimization, there is one I would find interesting, that would be to switch sp-state-machine `backend` trait to use `&mut ` on all accesses, thus removing the need for inner mutability (for trie cache and recorder).

Hmm.... is this actually going to affect the performance in a substantial way?

> > TrieDBIterator and TrieDBKeyIterator now use TrieDBRawIterator under the hood
> 
> wasn't it the case already?

Previously they had their dedicated `next` methods; now those were moved to the raw iterator, so now they're truly just a thin wrappers with no logic of their own.

cheme:
> Hmm.... is this actually going to affect the performance in a substantial way?

Not sure about perf, actually recorder and cache are already behind a refcell in the trie crate.
So probably the recorder is already accessed without lock, not sure about the local cache (may still be a useless rwlock there for the local caching, really not sure).

Would still make sense to me to just plain &mut ptr for read access operation and get rid of as much inner mutability as possible, but that is a different question.

koute:
@cheme Okay, I resurrected the fuzz tests (they didn't compile due to too old version of a dependency in `Cargo.toml`), renamed the `seek_prefix` to `seek`, made `next_raw_item` return references, and I've simplified the iterator methods a little more (I've reduced the indentation levels, removed unnecessary `expect`s, removed the extra internal `enum` in `next_raw_item`).

The logic's unchanged and I didn't touch the extra clones. You're right that we should be able to refactor it so that it appends to the `key_nibbles` directly, but from what I can see that'd be a more involved change so for now I left it as-is for a future PR.

cheme:
yes the fuzzer is quite often a bit behind, thanks.

> You're right that we should be able to refactor it so that it appends to the key_nibbles directly, but from what I can see that'd be a more involved change so for now I left it as-is for a future PR

I did check a bit, we probably want to return the key_nibbles after the partial (append the partial on `descend`), which seems fine to do, but at the same time the change will require to modify trie_codec.rs which can be a bit too many changes.

trie-db/src/triedb.rs
```diff
@@ -15,10 +15,10 @@
 #[cfg(feature = "std")]
 use crate::nibble::NibbleVec;
```
davxy, line 16:
```suggestion
use crate::{nibble::NibbleVec, node::Node};
```
Maybe we can group the stuff to be included with `std` under a single directive, thus also the `crate::rstd::{fmt, vec::Vec}` below

trie-db/src/triedb.rs
```diff
@@ -15,10 +15,10 @@
...
 use crate::{
-	iterator::TrieDBNodeIterator,
+	iterator::TrieDBRawIterator,
 	lookup::Lookup,
 	nibble::NibbleSlice,
-	node::{decode_hash, Node, NodeHandle, OwnedNode, Value},
+	node::{decode_hash, Node, NodeHandle, OwnedNode},
```
davxy, line 21:
```suggestion
	node::{decode_hash, NodeHandle, OwnedNode},
```

trie-db/src/iterator.rs
```diff
@@ -315,6 +321,241 @@ impl<'a, 'cache, L: TrieLayout> TrieDBNodeIterator<'a, 'cache, L> {
...
+	pub fn fetch_value(
+		&self,
+		key: &[u8],
+		prefix: Prefix,
+	) -> Result<DBValue, TrieHash<L>, CError<L>> {
+		TrieDBRawIterator::fetch_value(self.db, key, prefix)
+	}
+}
+
+impl<'a, 'cache, L: TrieLayout> TrieDBNodeIterator<'a, 'cache, L> {
```
davxy, line 542: why you used a separated `impl` block for the functions that follows?

trie-db/src/iterator.rs
```diff
@@ -315,6 +321,241 @@ impl<'a, 'cache, L: TrieLayout> TrieDBNodeIterator<'a, 'cache, L> {
...
+			}
+
+			return Some(Ok(key))
+		}
+		None
+	}
+}
+
+/// Iterator for going through all nodes in the trie in pre-order traversal order.
+pub struct TrieDBNodeIterator<'a, 'cache, L: TrieLayout> {
```
davxy, line 511: (Just curious about the design decision) why you decided to separate the `TrieDbNodeIterator` and `TriedDbRawIterator` and not embed everything in one single structure?

Intuition suggests that in this way you can export the raw iterator (`into_raw`) and reuse it to construct another iterator (`from_raw`). Maybe also you've preferred to keep the lower level component separated to have a more clean design?

davxy (approved):
Not a trie guru. But codewise looks good

trie-db/src/iterator.rs
koute: That's how it was originally. (:

But good point, we can merge those.

cheme:
Missed you by a few second, but trie-bench changelog can also be updated (not really important tbh), otherwhise latest changes still looks good.

koute:
> but trie-bench changelog can also be updated

Ah, sorry, I didn't realized that one's also released on crates.io. I'll update it.

PR #185, opened 2023-02-23
Return `Arc` instead of `Rc` from the iterators to make them `Send`

koute:
This PR replaces the `Rc` that's returned by the iterators with an `Arc`.

I need this to make the iterators `Send`, which I need for further performance improvements in `substrate`.

### DIFF
--- a/test-support/reference-trie/Cargo.toml
+++ b/test-support/reference-trie/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "reference-trie"
-version = "0.27.0"
+version = "0.28.0"
 authors = ["Parity Technologies <admin@parity.io>"]
 description = "Simple reference trie format"
 repository = "https://github.com/paritytech/trie/"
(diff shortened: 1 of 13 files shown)


altonen (approved):

cheme:
Does TrieDBNodeIterator need to be Send (if not, could also make send the other needed iterators as long as they cannot leek rc (the internal triedbnodeiterator))?

bkchr (approved):
A good old `is_sync` test maybe?

melekes (approved):
👍

koute:
> Does TrieDBNodeIterator need to be Send (if not, could also make send the other needed iterators as long as they cannot leek rc (the internal triedbnodeiterator))?

I just need `TrieDBRawIterator` to be `Send`, but since `TrieDBNodeIterator` is a thin wrapper around it it'll also automatically become `Send`. There are no other `Rc`s anywhere left so I think everything now will be `Send` anyway?

> A good old `is_sync` test maybe?

Good idea. I'll add it.

PR #187, opened 2023-03-07
remove hash of known value for leaf in compact proof.

cheme:
While working on  https://github.com/cheme/trie/tree/proof_size_tracking, I found that for leaf the hash of detached value was kept in proof.
This pr remove it.
On decode we still overwrite the hash the same way, so this should not need any specific client upgrade (could be an idea later to drop on hash present when it is included in next proof item).
Note that the code is correct for branch already.

### DIFF
--- a/trie-db/src/trie_codec.rs
+++ b/trie-db/src/trie_codec.rs
@@ -111,7 +111,16 @@ impl<C: NodeCodec> EncoderStackEntry<C> {
 			self.node.node_plan()
 		};
 		let mut encoded = match node_plan {
-			NodePlan::Empty | NodePlan::Leaf { .. } => node_data.to_vec(),
+			NodePlan::Empty => node_data.to_vec(),
+			NodePlan::Leaf { partial, value } => {
+				if self.omit_value {
+					let partial = partial.build(node_data);
+				let value = value.build(node_data);
+				C::leaf_node(partial.right_iter(), partial.len(), value)
+				} else {
+					node_data.to_vec()
+				}
+			},
 			NodePlan::Extension { partial, child: _ } =>
 				if !self.omit_children[0] {
 					node_data.to_vec()
--- a/trie-db/src/trie_codec.rs
+++ b/trie-db/src/trie_codec.rs
@@ -112,15 +112,14 @@ impl<C: NodeCodec> EncoderStackEntry<C> {
 		};
 		let mut encoded = match node_plan {
 			NodePlan::Empty => node_data.to_vec(),
-			NodePlan::Leaf { partial, value } => {
+			NodePlan::Leaf { partial, value } =>
 				if self.omit_value {
 					let partial = partial.build(node_data);
-				let value = value.build(node_data);
-				C::leaf_node(partial.right_iter(), partial.len(), value)
+					let value = value.build(node_data);
+					C::leaf_node(partial.right_iter(), partial.len(), value)
 				} else {
 					node_data.to_vec()
-				}
-			},
+				},
 			NodePlan::Extension { partial, child: _ } =>
 				if !self.omit_children[0] {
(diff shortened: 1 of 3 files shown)


trie-db/src/trie_codec.rs
```diff
@@ -111,7 +111,15 @@ impl<C: NodeCodec> EncoderStackEntry<C> {
 			self.node.node_plan()
 		};
 		let mut encoded = match node_plan {
-			NodePlan::Empty | NodePlan::Leaf { .. } => node_data.to_vec(),
+			NodePlan::Empty => node_data.to_vec(),
+			NodePlan::Leaf { partial, value } =>
+				if self.omit_value {
+					let partial = partial.build(node_data);
+					let value = value.build(node_data);
+					C::leaf_node(partial.right_iter(), partial.len(), value)
```
bkchr, line 119: Given the naming `omit_value` and then still seeing this here, I'm a little bit confused :P

bkchr (approved):
Can we get some test?

trie-db/src/trie_codec.rs
cheme: it s a 0 length value. Will rewrite that tomorrow to make it explicit.

trie-db/src/trie_codec.rs
```diff
@@ -99,24 +99,14 @@ impl<C: NodeCodec> EncoderStackEntry<C> {
...
+		let node_plan = self.node.node_plan();
 		let mut encoded = match node_plan {
 			NodePlan::Empty => node_data.to_vec(),
-			NodePlan::Leaf { partial, value } =>
+			NodePlan::Leaf { partial, value: _ } =>
 				if self.omit_value {
 					let partial = partial.build(node_data);
-					let value = value.build(node_data);
-					C::leaf_node(partial.right_iter(), partial.len(), value)
+					let no_value = crate::node::Value::Inline(&[]);
```
koute, line 108: You could put this in a const and use it in other branches, e.g.

```rust
const NO_VALUE: crate::node::Value<'static> = crate::node::Value::Inline(&[]);
```

trie-db/src/trie_codec.rs
```diff
@@ -128,17 +118,29 @@ impl<C: NodeCodec> EncoderStackEntry<C> {
 					let empty_child = ChildReference::Inline(C::HashOut::default(), 0);
 					C::extension_node(partial.right_iter(), partial.len(), empty_child)
 				},
-			NodePlan::Branch { value, children } => C::branch_node(
-				Self::branch_children(node_data, &children, &self.omit_children)?.iter(),
-				value.as_ref().map(|v| v.build(node_data)),
-			),
+			NodePlan::Branch { value, children } => {
+				let value = if self.omit_value {
+					value.as_ref().map(|_v| crate::node::Value::Inline(&[]))
```
koute, line 123: This could be simplified to:

```rust
value.is_some().then_some(NO_VALUE)
```

...not sure whether this is a little too much code golf-y or not though, so up to you. :P

koute (approved):

trie-db/src/trie_codec.rs
```diff
@@ -34,6 +34,8 @@ use crate::{
 };
 use hash_db::{HashDB, Prefix};
 
+const OMMIT_VALUE_HASH: crate::node::Value<'static> = crate::node::Value::Inline(&[]);
```
koute, line 37:
```suggestion
const OMIT_VALUE_HASH: crate::node::Value<'static> = crate::node::Value::Inline(&[]);
```

You got one `M` too many here. (:

cheme: Thanks :)

arkpar (approved):

PR #190, opened 2023-03-17
Fix `TrieDBRawIterator::prefix_then_seek`

koute:
After calling the `TrieDBRawIterator::prefix_then_seek` the iterator is not supposed to return any keys outside of the requested prefix; this unfortunately [is not the case](https://github.com/polkadot-js/apps/issues/9103), and this PR fixes the issue.

### Implementation notes

- Considering the complexity of the code I didn't completely trust myself to get this right, so I also added a fuzz test for this method. So far it's been running for quite a while and stopped barfing any new errors.
- For the fuzz test and for the tests I've copy-pasted the trie layout code from Substrate. The reference trie here already contains a Substrate-like trie layout, but it's not *exactly* the same. If I'm going to fuzz on something I'd rather do it on exactly the same thing we use in production.
- I only test and fuzz on the trie layout from Substrate. For historical reasons (e.g. we had to be Ethereum compatible in the past) the trie crate here is generic over various different layouts, but those 1) do behave differently (AFAIK some of the tests fail on other layouts), and 2) we really *don't* need to support those (honestly, I think we should just remove most of this flexibility as it just makes the code more complex for absolutely zero benefit to us).
- For the tests I've added a reproduction of the real-world issue that was affected by the bug this PR fixes, and also a few extra tests for anything that the fuzzer barfed while I was developing the fix.
- The case of what happens when both a start key and a prefix key is specified was not very well tested and it was mostly underspecified; now it is tested and behaves in a way that (I think) makes intuitive sense (take look at the `filter` in the fuzz test I added).
- Bumped the crate versions for immediate release.

### Potential future work

- Refactor the rest of the tests to use the `paste` crate and run each trie layout as their own test instead of running every layout inside of a single test.
- Refactor the rest of the fuzz tests to use the `arbitrary` crate instead of manually generating the fuzz input data.
- Cut down on the genericity and simplify everything.

### DIFF
--- a/test-support/reference-trie/src/substrate.rs
+++ b/test-support/reference-trie/src/substrate.rs
@@ -1,0 +1,725 @@
+// Copyright (C) Parity Technologies (UK) Ltd.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Codec and layout directly copy-pasted from substrate with minimal modifications.
+
+use core::{borrow::Borrow, iter::once, marker::PhantomData, ops::Range};
+use hash_db::Hasher;
+use parity_scale_codec as codec;
+use parity_scale_codec::{Compact, Decode, Encode, Input, Output};
+use trie_db::{
+	nibble_ops,
+	node::{NibbleSlicePlan, NodeHandlePlan, NodePlan, Value, ValuePlan},
+	ChildReference, NodeCodec as NodeCodecT, TrieConfiguration, TrieLayout,
+};
+
+/// Constants used into trie simplification codec.
+mod trie_constants {
+	const FIRST_PREFIX: u8 = 0b_00 << 6;
+	pub const LEAF_PREFIX_MASK: u8 = 0b_01 << 6;
+	pub const BRANCH_WITHOUT_MASK: u8 = 0b_10 << 6;
+	pub const BRANCH_WITH_MASK: u8 = 0b_11 << 6;
+	pub const EMPTY_TRIE: u8 = FIRST_PREFIX | (0b_00 << 4);
+	pub const ALT_HASHING_LEAF_PREFIX_MASK: u8 = FIRST_PREFIX | (0b_1 << 5);
+	pub const ALT_HASHING_BRANCH_WITH_MASK: u8 = FIRST_PREFIX | (0b_01 << 4);
+	pub const ESCAPE_COMPACT_HEADER: u8 = EMPTY_TRIE | 0b_00_01;
+}
+
+pub const TRIE_VALUE_NODE_THRESHOLD: u32 = 33;
+
+/// Codec-flavored TrieStream.
+#[derive(Default, Clone)]
+pub struct TrieStream {
+	/// Current node buffer.
+	buffer: Vec<u8>,
+}
+
+fn branch_node_bit_mask(has_children: impl Iterator<Item = bool>) -> (u8, u8) {
+	let mut bitmap: u16 = 0;
+	let mut cursor: u16 = 1;
+	for v in has_children {
+		if v {
+			bitmap |= cursor
+		}
+		cursor <<= 1;
+	}
+	((bitmap % 256) as u8, (bitmap / 256) as u8)
+}
+
+/// Create a leaf/branch node, encoding a number of nibbles.
+fn fuse_nibbles_node(nibbles: &[u8], kind: NodeKind) -> impl Iterator<Item = u8> + '_ {
+	let size = nibbles.len();
+	let iter_start = match kind {
+		NodeKind::Leaf => size_and_prefix_iterator(size, trie_constants::LEAF_PREFIX_MASK, 2),
+		NodeKind::BranchNoValue =>
+			size_and_prefix_iterator(size, trie_constants::BRANCH_WITHOUT_MASK, 2),
+		NodeKind::BranchWithValue =>
+			size_and_prefix_iterator(size, trie_constants::BRANCH_WITH_MASK, 2),
+		NodeKind::HashedValueLeaf =>
+			size_and_prefix_iterator(size, trie_constants::ALT_HASHING_LEAF_PREFIX_MASK, 3),
+		NodeKind::HashedValueBranch =>
+			size_and_prefix_iterator(size, trie_constants::ALT_HASHING_BRANCH_WITH_MASK, 4),
+	};
+	iter_start
+		.chain(if nibbles.len() % 2 == 1 { Some(nibbles[0]) } else { None })
+		.chain(nibbles[nibbles.len() % 2..].chunks(2).map(|ch| ch[0] << 4 | ch[1]))
+}
+
+use trie_root::Value as TrieStreamValue;
+impl trie_root::TrieStream for TrieStream {
+	fn new() -> Self {
+		Self { buffer: Vec::new() }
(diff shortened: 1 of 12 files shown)


trie-db/src/iterator.rs
```diff
@@ -288,35 +288,60 @@ impl<L: TrieLayout> TrieDBRawIterator<L> {
...
-				if len > prefix_len {
-					self.trail = self.trail.split_off(i);
-					return Ok(())
-				}
+		if prefix.is_empty() {
+			// There's no prefix, so just seek.
+			return self.seek(db, seek).map(|_| ())
+		}
+
+		if seek.is_empty() || seek <= prefix {
```
cheme, line 296: +1 for this new corner case

test-support/reference-trie/src/substrate.rs
```diff
@@ -0,0 +1,725 @@
+// Copyright (C) Parity Technologies (UK) Ltd.
```
cheme, line 1: Should remove test-support/reference-trie/src/substrate-like.rs (can be in a second time, might be some slight difference??).

trie-db/src/iterator.rs
```diff
@@ -288,35 +288,60 @@ impl<L: TrieLayout> TrieDBRawIterator<L> {
...
+				NodePlan::Empty => {},
+				NodePlan::Branch { .. } => {
+					len += 1;
+				},
+				NodePlan::Leaf { partial, .. } => {
+					len += partial.len();
+				},
+				NodePlan::Extension { partial, .. } => {
+					len += partial.len();
+				},
```
bkchr, line 333: @cheme why are we not adding here `len += 1`? What is the difference to the others?

trie-db/src/iterator.rs
```diff
@@ -288,35 +288,60 @@ impl<L: TrieLayout> TrieDBRawIterator<L> {
...
+			// so just directly go to the prefix.
+			return self.prefix(db, prefix)
+		}
+
+		if !seek.starts_with(prefix) {
+			// We're supposed to seek *after* the prefix,
+			// so just return an empty iterator.
+			self.trail.clear();
+			return Ok(())
+		}
```
bkchr, line 308: This looks like something that shouldn't be supported at all and the upperlying logic should forbid this?

trie-db/src/iterator.rs
```diff
@@ -288,35 +288,60 @@ impl<L: TrieLayout> TrieDBRawIterator<L> {
...
+
+		if !self.seek(db, prefix)? {
```
cheme, line 310: So the fix is to seek in two steps (and avoid doing it twice if prefix == seek actually just using prefix) with second seek that cannot go outside the prefix due to prechecks.

Indeed my code was broken on empty prefix, seek going after the prefix and trail being clean as if it is into the prefix (original code did not test the return value).

cheme (approved):
Thanks a lot for fixing this one (I did not go through the fuzz but prefer to approve asap).

> Refactor the rest of the fuzz tests to use the arbitrary crate instead of manually generating the fuzz input data.

sure is desirable.

> Cut down on the genericity and simplify everything.

we should fork to have substrate only crate, this is something we want for a while.
my opinion was to wait for removing rocksdb support (which will also allow simplifying a lot), but probably a better idea to
do sooner.

trie-db/src/iterator.rs
```diff
@@ -288,35 +288,60 @@ impl<L: TrieLayout> TrieDBRawIterator<L> {
...
+			// The database doesn't have a key with such a prefix.
+			self.trail.clear();
+			return Ok(())
+		}
```
bkchr, line 314: This is the actual fix or?

bkchr: Because before we still returned some data?

cheme:
From what I read this bug is not in a host function (only clear_prefix v3 can go in the code path but it is "register_only" cc @bkchr is it possible to still use the function? (in this case I am not sure if it can be an issue since runtime will only call it with a cursor if it is not empty, and having cursor at last position do not make sense as we know it is the last position)).

bkchr (approved):
I also checked Substrate and the good thing is that we don't have the `clear_prefix` host functions with `maybe_cursor` enabled yet. So, we should not be able to have build a block with this bug? @cheme and @koute please check again that I'm right!

bkchr:
> From what I read this bug is not in a host function (only clear_prefix v3 can go in the code path but it is "register_only" cc @bkchr is it possible to still use the function? (in this case I am not sure if it can be an issue since runtime will only call it with a cursor if it is not empty, and having cursor at last position do not make sense as we know it is the last position)).

Ahh you also came to the same conclusion as me! Good :D

trie-db/src/iterator.rs
cheme: Branches contains a nibble that is not part of the nible slice, the index to the child node.
a trie with key (nibbles not bytes) '1,0' and '1,2' will be a branch and two child with empty partial key. The partial key of the branch will be 1 and the branch contains 0 and 2 nibble as index to the leafs.

test-support/reference-trie/src/substrate.rs
koute: Yeah, I think we can remove that one if we're going to have the real thing. I just didn't want to needlessly bloat up this PR with more changes.

koute:
> I also checked Substrate and the good thing is that we don't have the `clear_prefix` host functions with `maybe_cursor` enabled yet. So, we should not be able to have build a block with this bug? @cheme and @koute please check again that I'm right!

From a cursory look this sounds about right.

trie-db/src/iterator.rs
koute: > This is the actual fix or?
>
> Because before we still returned some data?

I think @cheme already explained it pretty well [here](https://github.com/paritytech/trie/pull/190#discussion_r1140112435), but basically yes, previously we seeked to a key which might not start with the prefix anymore, and the code just checked that the length is good but didn't check that it's still within the prefix.

koute:
Okay, since I want to put up a substrate PR with this fix I'll merge this now. If you'd still like for me to make any further changes feel free to leave them and I'll do them in a followup.

trie-db/src/iterator.rs
koute: > I was thinking about it from the sp_io perspective and api should be I think:
> 
>     * prefix some prefix
> 
>     * seek some seek key without the prefix
> 
> 
> So the first value you access will not be `seek` when into the prefix, but `prefix ++ seek`

So if I understand this right, are you suggesting that the subsequent seek after seeking to the prefix should be allowed to "break out" of the prefix and iterate over values which do not start with the prefix?

cheme: It is more an api suggestion that would avoid some corner cases (current code looks plain good to me):
Instead of calling `prefix_then_seek(b"prefix", b"prefix_seek")` or `prefix_then_seek(b"prefix", b"not_in_prefix")`.
We will call `prefix_then_seek(b"prefix", b"_seek")` and could not do the second call at all.

cheme: But doing this change is breaking api. I think it would be more interesting at sp_io level also.
(was just a thinking I had during lunch and wanted to share).

koute: Hmmm... I guess we could do that, but that might be more awkward to use depending on the use case I guess? But then it could be easier in some other cases? And also, in this case I think it could make sense for it to return keys which are also have the prefix stripped, e.g. assume the following code:

```rust
loop {
    let accounts = fetch_values(ACCOUNTS_PREFIX, cursor, 100);
    // ...
    cursor = match accounts.last() {
        Some(account) => account,
        None => break
    }
}
```

For this code to work and be convenient to use it'll have to return the same "type" of key (stripped of the prefix or not stripped) as what the function accepts in `cursor`, otherwise you'll have to do awkward concats or have to slice off the prefix yourself.

---

Also, only tangentially related, but maybe naming wise we could rename `prefix` to `only_starting_with` and `seek` to `skip_until_or_after` or something like that? It's a little wordy but it would explain what it does more clearly.

cheme: Thinking twice could be good for trie to have better naming, but I got pretty use to the current one so my opinion is rather biased.

trie-db/src/iterator.rs
bkchr: Ahh okay!

PR #194, opened 2023-06-02
Make `trie_nodes_recorded_for_key` work for inline values

bkchr:
`trie_nodes_recorded_for_key` was not working properly for inline values. It would always return `RecordedForKey::None` while we actually have accessed and recorded all the trie nodes for the value. The pr introduces `TrieAccess::InlineValue` to communicate this access to the recorder properly to make it then return `RecordedForKey::Value`.

### DIFF
--- a/trie-db/src/lookup.rs
+++ b/trie-db/src/lookup.rs
@@ -58,7 +58,13 @@ where
 		query: Q,
 	) -> Result<Q::Item, TrieHash<L>, CError<L>> {
 		match v {
-			Value::Inline(value) => Ok(query.decode(&value)),
+			Value::Inline(value) => {
+				if let Some(recorder) = recorder {
+					recorder.record(TrieAccess::InlineValue { full_key });
+				}
+
+				Ok(query.decode(&value))
+			},
 			Value::Node(hash) => {
 				let mut res = TrieHash::<L>::default();
 				res.as_mut().copy_from_slice(hash);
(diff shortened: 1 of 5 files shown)


trie-db/src/lookup.rs
```diff
@@ -167,7 +179,13 @@ where
 				full_key,
 				|v, _, full_key, _, recorder, _| {
 					Ok(match v {
-						Value::Inline(v) => L::Hash::hash(&v),
+						Value::Inline(v) => {
+							if let Some(recoder) = recorder.as_mut() {
+								recoder.record(TrieAccess::InlineValue { full_key });
```
cheme, line 184: Maybe a comment to indicate we prefer recording a value access than a hash here.

trie-db/src/lookup.rs
```diff
@@ -211,7 +229,13 @@ where
 				full_key,
 				cache,
 				|value, _, full_key, _, _, recorder| match value {
-					ValueOwned::Inline(value, hash) => Ok((hash, Some(value.clone()))),
+					ValueOwned::Inline(value, hash) => {
```
cheme, line 232: likewise, should we record TrieAccess::InlineHash? Probably not, but maybe a comment.

cheme: function `look_up_with_cache` is recording Value access when non inline, it can probably also do for the inline case.

arkpar:
@bkchr Is this still relevane? Why do we need to record inline nodes in the first place?

bkchr:
> @bkchr Is this still relevane? Why do we need to record inline nodes in the first place?

This is just about telling the recorder that we have seen the value. This is required when we use the cache, to be able to fetch the value from the cache. When the recorder and cache are enabled, we only fetch a value from the cache if we have recorded all the trie nodes up to this value.

trie-db/src/lookup.rs
cheme: > function look_up_with_cache is recording Value access when non inline, it can probably also do for the inline case.

Not really sure what I meant here at the time, make only sense if there was node value in inline node (which is doable but not really a sensible configuration).

cheme (approved):

arkpar (approved):

PR #196, opened 2023-06-30
Query plan proofs

cheme:
This branch adds method to create a proof form a set of key value request (following https://github.com/w3f/PPPs/pull/10).

Proof can be split depending on size limits.
To create a splitted proof there is two mode:
- statefull where the same state need to be use for every query.
- stateless every part of the proof can be produce from parameters without state (this is more complex but is what should be use in substrate rpc).

Verification is only statefull (for a client there is no sense in obtaining incomplete proof without having all the other part of proofs).

It propose three kind of proofs, only full nodes is needed and there is chance I remove the Content proof when switching this pr out of draft.

The substrate branch using it is https://github.com/cheme/substrate/tree/trie_hash (note that no limit is set by default, but this should be configurable for at node start).


cheme:
PR should be ready for review, I did remove the content kind of proof.  If it is still a bit too big the compact version could also be removed (yet I think it is very nice to have).

PR #199, opened 2023-08-21
trie-db: Fetch the closest merkle value

lexnv:
This PR adds support for fetching the merkle value directly from the trie-db.

- Adds `get_closest_merkle_value` on the `Trie` trait
- `TrieDb` iterates over the nibbles of the provided key
  - At each iteration DB is queried for the containing node
  - last value encountered on the path from the root to the key is returned
- During trie traversal, the internal recorder is updated with trie actions
  - This should make other trie queries to the key more efficient
- Test with `HashedValueNoExtThreshold<1>` as trie layout and value update at given key

We could build on this PR to have support for recorded Merkel Value in the cache.

Would love to get your thoughts on this @cheme @arkpar 🙏 
(For reference @tomaka to ensure we don't break anything wrt to compatibilities between smoldot and substrate)

Part of https://github.com/paritytech/substrate/issues/14550.

// @paritytech/subxt-team

### DIFF
--- a/trie-db/src/lib.rs
+++ b/trie-db/src/lib.rs
@@ -277,6 +277,12 @@ pub trait Trie<L: TrieLayout> {
 		query: Q,
 	) -> Result<Option<Q::Item>, TrieHash<L>, CError<L>>;
 
+	/// Returns the merkle value of the closest descendant node of the key.
+	fn get_closest_merkle_value(
+		&self,
+		key: &[u8],
+	) -> Result<Option<TrieHash<L>>, TrieHash<L>, CError<L>>;
+
 	/// Returns a depth-first iterator over the elements of trie.
 	fn iter<'a>(
 		&'a self,
@@ -402,6 +408,13 @@ impl<'db, 'cache, L: TrieLayout> Trie<L> for TrieKinds<'db, 'cache, L> {
 		query: Q,
 	) -> Result<Option<Q::Item>, TrieHash<L>, CError<L>> {
 		wrapper!(self, get_with, key, query)
+	}
+
+	fn get_closest_merkle_value(
+		&self,
+		key: &[u8],
+	) -> Result<Option<TrieHash<L>>, TrieHash<L>, CError<L>> {
+		wrapper!(self, get_closest_merkle_value, key)
 	}
 
 	fn iter<'a>(
--- a/trie-db/src/lib.rs
+++ b/trie-db/src/lib.rs
@@ -277,8 +277,13 @@ pub trait Trie<L: TrieLayout> {
 		query: Q,
 	) -> Result<Option<Q::Item>, TrieHash<L>, CError<L>>;
 
-	/// Returns the merkle value of the closest descendant node of the key.
-	fn get_closest_merkle_value(
+	/// Look up the merkle value of the node that is the closest descendant for the provided
+	/// key.
+	///
+	/// When the provided key leads to a node, then the merkle value of that node
+	/// is returned. However, if the key does not lead to a node, then the merkle value
+	/// of the closest descendant is returned. `None` if no such descendant exists.
+	fn lookup_first_descendant(
 		&self,
 		key: &[u8],
 	) -> Result<Option<TrieHash<L>>, TrieHash<L>, CError<L>>;
@@ -410,11 +415,11 @@ impl<'db, 'cache, L: TrieLayout> Trie<L> for TrieKinds<'db, 'cache, L> {
 		wrapper!(self, get_with, key, query)
 	}
 
-	fn get_closest_merkle_value(
+	fn lookup_first_descendant(
 		&self,
 		key: &[u8],
 	) -> Result<Option<TrieHash<L>>, TrieHash<L>, CError<L>> {
-		wrapper!(self, get_closest_merkle_value, key)
+		wrapper!(self, lookup_first_descendant, key)
 	}
 
 	fn iter<'a>(
--- a/trie-db/src/lib.rs
+++ b/trie-db/src/lib.rs
@@ -286,7 +286,7 @@ pub trait Trie<L: TrieLayout> {
 	fn lookup_first_descendant(
 		&self,
 		key: &[u8],
-	) -> Result<Option<TrieHash<L>>, TrieHash<L>, CError<L>>;
+	) -> Result<Option<MerkleValue<TrieHash<L>>>, TrieHash<L>, CError<L>>;
 
 	/// Returns a depth-first iterator over the elements of trie.
 	fn iter<'a>(
@@ -418,7 +418,7 @@ impl<'db, 'cache, L: TrieLayout> Trie<L> for TrieKinds<'db, 'cache, L> {
 	fn lookup_first_descendant(
 		&self,
 		key: &[u8],
-	) -> Result<Option<TrieHash<L>>, TrieHash<L>, CError<L>> {
+	) -> Result<Option<MerkleValue<TrieHash<L>>>, TrieHash<L>, CError<L>> {
 		wrapper!(self, lookup_first_descendant, key)
 	}
 
@@ -755,4 +755,16 @@ impl From<Bytes> for BytesWeak {
 	fn from(bytes: Bytes) -> Self {
 		Self(rstd::sync::Arc::downgrade(&bytes.0))
 	}
+}
+
+/// A value returned by [`Trie::lookup_first_descendant`].
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum MerkleValue<H> {
+	/// The merkle value is the node data itself when the
+	/// node data is smaller than `MAX_INLINE_VALUE`.
+	///
+	/// Note: The case of inline nodes.
+	Node(Vec<u8>),
+	/// The merkle value is the hash of the node.
+	Hash(H),
 }
--- a/trie-db/src/lib.rs
+++ b/trie-db/src/lib.rs
@@ -757,7 +757,9 @@ impl From<Bytes> for BytesWeak {
 	}
 }
 
-/// A value returned by [`Trie::lookup_first_descendant`].
+/// Either the `hash` or `value` of a node depending on its size.
+///
+/// If the size of the node `value` is bigger or equal than `MAX_INLINE_VALUE` the `hash` is returned.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum MerkleValue<H> {
 	/// The merkle value is the node data itself when the
--- a/trie-db/src/lib.rs
+++ b/trie-db/src/lib.rs
@@ -277,7 +277,7 @@ pub trait Trie<L: TrieLayout> {
 		query: Q,
 	) -> Result<Option<Q::Item>, TrieHash<L>, CError<L>>;
 
-	/// Look up the merkle value of the node that is the closest descendant for the provided
+	/// Look up the [`MerkleValue`] of the node that is the closest descendant for the provided
 	/// key.
 	///
 	/// When the provided key leads to a node, then the merkle value of that node
--- a/trie-db/src/lib.rs
+++ b/trie-db/src/lib.rs
@@ -759,7 +759,8 @@ impl From<Bytes> for BytesWeak {
 
 /// Either the `hash` or `value` of a node depending on its size.
 ///
-/// If the size of the node `value` is bigger or equal than `MAX_INLINE_VALUE` the `hash` is returned.
+/// If the size of the node `value` is bigger or equal than `MAX_INLINE_VALUE` the `hash` is
+/// returned.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum MerkleValue<H> {
 	/// The merkle value is the node data itself when the
--- a/trie-db/src/lookup.rs
+++ b/trie-db/src/lookup.rs
@@ -19,8 +19,8 @@ use crate::{
 	node::{decode_hash, Node, NodeHandle, NodeHandleOwned, NodeOwned, Value, ValueOwned},
 	node_codec::NodeCodec,
 	rstd::{boxed::Box, vec::Vec},
-	Bytes, CError, CachedValue, DBValue, Query, RecordedForKey, Result, TrieAccess, TrieCache,
-	TrieError, TrieHash, TrieLayout, TrieRecorder,
+	Bytes, CError, CachedValue, DBValue, MerkleValue, Query, RecordedForKey, Result, TrieAccess,
+	TrieCache, TrieError, TrieHash, TrieLayout, TrieRecorder,
 };
 use hash_db::{HashDBRef, Hasher, Prefix};
 
@@ -143,7 +143,7 @@ where
 		self,
 		full_key: &[u8],
 		nibble_key: NibbleSlice,
-	) -> Result<Option<TrieHash<L>>, TrieHash<L>, CError<L>> {
+	) -> Result<Option<MerkleValue<TrieHash<L>>>, TrieHash<L>, CError<L>> {
 		self.inner_lookup_first_descendent(nibble_key, full_key)
 	}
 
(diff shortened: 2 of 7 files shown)


trie-db/src/lookup.rs
```diff
@@ -655,4 +668,133 @@ where
 		}
 		Ok(None)
 	}
+
+	/// Look up the closest merkle value of the provided key.
+	///
+	/// When the provided key leads to a node, then the merkle value of that node
+	/// is returned. However, if the key does not lead to a node, then the closest
+	/// merkle value is returned.
+	fn look_up_merkle_without_cache(
```
arkpar, line 677: This should be called `lookup_first_descendent_without_cache` or something like this this.

trie-db/src/lookup.rs
```diff
@@ -655,4 +668,133 @@ where
 		}
 		Ok(None)
 	}
+
+	/// Look up the closest merkle value of the provided key.
```
arkpar, line 672:
```suggestion
	/// Look up the merkle value (hash) of the node that is the closest descendant for the provided key.
```

trie-db/src/lookup.rs
```diff
@@ -655,4 +668,133 @@ where
...
+	///
+	/// When the provided key leads to a node, then the merkle value of that node
+	/// is returned. However, if the key does not lead to a node, then the closest
+	/// merkle value is returned.
```
arkpar, line 676:
```suggestion
	/// When the provided key leads to a node, then the merkle value of that node
	/// is returned. However, if the key does not lead to a node, then the merkle value
	/// of the closes descendant is return. `None` if no such descendant exists.
```

trie-db/test/src/triedb.rs
```diff
@@ -644,6 +644,80 @@ fn test_recorder_with_cache_get_hash_internal<T: TrieLayout>() {
 	}
 }
 
+#[test]
+fn test_merkle_value() {
```
arkpar, line 648: The test should also check that the functions works correctly for keys that aren't acuatlly in the tree.
E.g. for a tree with nodes ["AAA"] it should return a closest descendant for the key "AA" but not for "BB". For the tree with nodes ["AAAA", "AABA] closes descendant for the key "AA" should be the branch node, etc. 
Also there's no need to make changes to the test trie after it is created really.


Also, you don't need to

jsdw: > Also, you don't need to

Argh, the suspense :D

trie-db/src/lookup.rs
lexnv: That makes sense, thanks for the suggestion!

trie-db/test/src/triedb.rs
```diff
@@ -644,6 +644,173 @@ fn test_recorder_with_cache_get_hash_internal<T: TrieLayout>() {
...
+	// Data set.
+	let key_value = vec![
+		(b"A".to_vec(), vec![1; 64]),
+		(b"AA".to_vec(), vec![2; 64]),
+		(b"AAAA".to_vec(), vec![3; 64]),
+		(b"AAB".to_vec(), vec![4; 64]),
+		(b"AABBBB".to_vec(), vec![4; 64]),
+		(b"AB".to_vec(), vec![5; 4]),
+		(b"B".to_vec(), vec![6; 64]),
+	];
```
cheme, line 661: Should be good to test with vec of len 1 so we also query inline nodes.

trie-db/src/lookup.rs
```diff
@@ -655,4 +668,175 @@ where
...
+						let common_prefix_len = partial.common_prefix(&slice);
+
+						let start_matches = if partial.len() <= slice.len() {
+							// Extension slice can be longer than remainder of the provided key
+							// (descendent), ensure the extension slice starts with the remainder
+							// of the provided key.
+							//
+							// This effectively returns the hash of the `Node::Branch` at the next
+							// iteration.
+							common_prefix_len == partial.len()
```
cheme, line 737: Then in case the key is in the middle of the extension partial and match all along, the return merkle value should be the one of the extension.
But tbh, I would just put all `extension` related as unimplemented (and remove the variant from the tests).

trie-db/src/lookup.rs
```diff
@@ -655,4 +668,175 @@ where
...
+						if partial.len() != slice.len() {
+							self.record(|| TrieAccess::NonExisting { full_key });
+						}
+
+						return Ok(Some(hash))
+					},
+					Node::Extension(slice, item) => {
+						let common_prefix_len = partial.common_prefix(&slice);
+
+						let start_matches = if partial.len() <= slice.len() {
```
cheme, line 730: Should be strictly inferior here.

trie-db/src/lookup.rs
```diff
@@ -655,4 +668,175 @@ where
...
+								return Ok(Some(hash))
+							} else {
+								return Ok(None)
+							}
+						}
+
+						// Partial key is longer or equal than the branch slice.
+						// Ensure partial key starts with the branch slice.
+						if common_prefix_len != slice.len() {
+							self.record(|| TrieAccess::NonExisting { full_key });
```
cheme, line 795: what if common_prefix_len is less than slice.len() and at the same time common_prefix_len is equal to partial.len().
Aka the queried key is into the partial, then I think we need to return the hash of the current branch.
(similar case as starts_with in leaf).

trie-db/src/lookup.rs
```diff
@@ -655,4 +668,175 @@ where
...
+
+									return Ok(None)
+								},
+							}
+						}
+					},
+					Node::Empty => {
+						self.record(|| TrieAccess::NonExisting { full_key });
+
+						return Ok(Some(hash))
```
cheme, line 823: I would return Ok(None) (rational being that the queried key is not contained in the trie).

trie-db/src/lookup.rs
```diff
@@ -655,4 +668,175 @@ where
...
+
+				// check if new node data is inline or hash.
+				match next_node {
+					NodeHandle::Hash(data) => {
+						hash = decode_hash::<L::Hash>(data)
+							.ok_or_else(|| Box::new(TrieError::InvalidHash(hash, data.to_vec())))?;
+						break
+					},
+					NodeHandle::Inline(data) => {
+						node_data = data;
```
cheme, line 835: I am not sure this would work (I think we will loop on querying the same node again).
What I would do with inline node:
- not fetch new node but still descend in node
- return hash(node_data) if we are in an inline node instead of hash.

cheme (commented):
@lexnv , thanks for the work, I got a few question on corner cases, notably inline node (I am not 100% sure, depends of how the spec is understood).

trie-db/src/lookup.rs
lexnv: Would this now be the case of `partial.len() < slice.len() { if slice.starts_with(&partial) } `?

Have changed the code to use `starts_with` instead of the common length prefix, which makes it easier to read for me, thanks!

trie-db/test/src/triedb.rs
```diff
@@ -644,6 +644,173 @@ fn test_recorder_with_cache_get_hash_internal<T: TrieLayout>() {
...
+	let expected = trie.get_closest_merkle_value(b"AABBBB").unwrap().unwrap();
+	assert_eq!(hash, expected);
+
+	// Prefix AABB in between AAB and AABBBB, but has different ending char.
+	let hash = trie.get_closest_merkle_value(b"AABBX").unwrap();
+	assert!(hash.is_none());
+}
+
+test_layouts!(test_merkle_value_single_key, test_merkle_value_single_key_internal);
+fn test_merkle_value_single_key_internal<T: TrieLayout>() {
```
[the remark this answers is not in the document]
lexnv: That's my understanding also, have added a test for it, thanks for the suggestion

trie-db/src/lookup.rs
lexnv: That makes sense! Have changed the code to also use `starts_with`.
Have kept the `extension` for not exposing a variation of `test_layouts!` but could place this under an `unimplemented` as well, let me know which one you think is better here 🙏

trie-db/src/lookup.rs
lexnv: IIUC for the `inline` case we could do something similar to:

```rust
NodeHandle::Inline(data) => {
    // Decode to ensure we are not following an empty node
    let decoded = match L::Codec::decode(data) {
	    Ok(node) => node,
	    Err(e) => return Err(Box::new(TrieError::DecoderError(hash, e))),
    };
    return match decoded {
	    Node::Empty => Ok(None),
	    _ => Ok(Some(hash)),
    }
},
```

With this change applied, the `test_merkle_value` seems to fail since we inlined

```
(b"AABBBB".to_vec(), vec![4; 1]),
(b"AB".to_vec(), vec![5; 1]),
(b"B".to_vec(), vec![6; 1]),
```

The failure is happening for key `let hash = trie.get_closest_merkle_value(b"AABC").unwrap();` which should not be present. Let me know if I got this right, thanks!

trie-db/src/lookup.rs
```diff
@@ -134,6 +134,19 @@ where
 		}
 	}
 
+	/// Look up the closest merkle value.
+	///
+	/// When the provided key leads to a node, then the merkle value of that node
+	/// is returned. However, if the key does not lead to a node, then the closest
+	/// merkle value is returned.
+	pub fn look_up_merkle(
```
arkpar, line 142: Should be called `lookup_first_descendant`

trie-db/src/lookup.rs
```diff
@@ -655,4 +668,174 @@ where
...
+	/// key.
+	///
+	/// When the provided key leads to a node, then the merkle value of that node
+	/// is returned. However, if the key does not lead to a node, then the merkle value
+	/// of the closest descendant is returned. `None` if no such descendant exists.
+	fn lookup_first_descendent_without_cache(
+		mut self,
+		nibble_key: NibbleSlice,
+		full_key: &[u8],
+	) -> Result<Option<TrieHash<L>>, TrieHash<L>, CError<L>> {
```
bkchr, line 682: How can this return `None`? If there doesn't exist any branch of the key, we should return the storage root?

bkchr: Forget what I said, I now realized the change to "closest descendant"  :see_no_evil:

trie-db/src/lookup.rs
```diff
@@ -655,4 +668,175 @@ where
...
+		let mut partial = nibble_key;
+		let mut hash = self.hash;
+		let mut key_nibbles = 0;
+
+		// this loop iterates through non-inline nodes.
+		for depth in 0.. {
+			let node_data = match self.db.get(&hash, nibble_key.mid(key_nibbles).left()) {
```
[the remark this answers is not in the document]
lexnv: That makes sense! Have written the code to support both pathways:
- I've used the cache if it exits
- If we don't have any caching enabled for the query; then I've transformed the decoded node into `NodeOwned`

Since we are now working with a partial NibbleSlice and a NodeOwned::NibbleVec, I've also added a `starts_with_slice` method (since the NibbleVec could have a padding) 

Let me know what you think of it, thanks! 🙏

cheme (approved):
Looks good.
Worth remembering that if at some point this get used in runtime (especially by a host function), not simply rpc, it would be worth fuzzing a bit.

arkpar (approved):

trie-db/src/lookup.rs
```diff
@@ -655,4 +668,219 @@ where
...
+			} else {
+				_owned_node = get_owned_node(depth)?;
+
+				self.record(|| TrieAccess::EncodedNode {
+					hash,
+					encoded_node: node_data.as_slice().into(),
+				});
+
+				&_owned_node
+			};
```
bkchr, line 738: Now we always need to create an owned node, which is a little bit shitty. But fine for now I think.

trie-db/src/lookup.rs
```diff
@@ -655,4 +668,219 @@ where
...
+	/// key.
+	///
+	/// When the provided key leads to a node, then the merkle value of that node
+	/// is returned. However, if the key does not lead to a node, then the merkle value
+	/// of the closest descendant is returned. `None` if no such descendant exists.
+	fn inner_lookup_first_descendent(
+		mut self,
+		nibble_key: NibbleSlice,
+		full_key: &[u8],
+	) -> Result<Option<TrieHash<L>>, TrieHash<L>, CError<L>> {
```
bkchr, line 682: The return type is wrong.
```
enum MerkleValue {
     /// Is returned if `v.len() < MAX_INLINE_VALUE`.
     Node(Vec<u8>),
     Hash(H),
}
```

bkchr (commented):
Left mainly some nitpicks. However, the return value is wrong. The merkle value is either the node data or the hash, depending on if this is an inline node or not.

trie-db/src/lookup.rs
cheme: I did propose to use hash for inline node (generally I would prefer if inline node is something users don't need to be aware off). But I did not double check the rpc proposal.

cheme: Maybe trie crate api could return this enum, but the rpc return hash of the data value for inline node (this way we can avoid some hashing if this api get use by runtime in the future).

bkchr: But the spec is speaking about the merkle value and not the hash. The point being that if the data is smaller, we save bandwidth etc. 



> (this way we can avoid some hashing if this api get use by runtime in the future)

This doesn't add any extra hashing. If you need hashing for the runtime, we can just do it where we use it. There would be no extra hashing involved as we currently also hash the inline node manually.

trie-db/src/lib.rs
```diff
@@ -277,6 +277,12 @@ pub trait Trie<L: TrieLayout> {
 		query: Q,
 	) -> Result<Option<Q::Item>, TrieHash<L>, CError<L>>;
 
+	/// Returns the merkle value of the closest descendant node of the key.
+	fn get_closest_merkle_value(
```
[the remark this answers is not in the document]
lexnv: Thanks! I've renamed this and added the comments from the `Lookup`

trie-db/src/sectriedb.rs
```diff
@@ -75,6 +75,13 @@ where
 		self.raw.get_with(L::Hash::hash(key).as_ref(), query)
 	}
 
+	fn get_closest_merkle_value(
+		&self,
+		_key: &[u8],
+	) -> Result<Option<TrieHash<L>>, TrieHash<L>, CError<L>> {
+		Ok(None)
```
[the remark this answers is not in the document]
lexnv: Thanks! I've forwarded to sectriedb and fatdb 🙏

trie-db/src/lookup.rs
lexnv: That makes sense! I've added the `MerkleValue` enum to be explicit about what the trie-db returns.

Then in substrate's RPC layer we could further `hash(MerkleValue::Node(..))` such that we return a consistent view of nodes and not care about inline ones. And if this function gets ever called from the runtime we save a `Layout:hash()` call 🙏

bkchr: > Then in substrate's RPC layer we could further `hash(MerkleValue::Node(..))` such that we return a consistent view of nodes and not care about inline ones.

This still goes against the size argument I laid out above. I opened https://github.com/paritytech/json-rpc-interface-spec/issues/88 to be sure that we use the correct value in the RPC.

trie-db/src/lib.rs
```diff
@@ -738,3 +756,15 @@ impl From<Bytes> for BytesWeak {
 		Self(rstd::sync::Arc::downgrade(&bytes.0))
 	}
 }
+
+/// A value returned by [`Trie::lookup_first_descendant`].
```
bkchr, line 760:
```suggestion
/// Either the `hash` or `value` of a node depending on its size.
///
/// If the size of the node `value` is bigger or equal than `MAX_INLINE_VALUE` the `hash` is returned.
```

bkchr (approved):

PR #202, opened 2023-09-13
Fix recorder for `NonExisting` keys

bkchr:
We handle `NonExisting` keys as if we have recorded the value. The reason behind this is that we have recorded all the trie nodes to proof that the key doesn't exist in the `trie`. So, next time we want to access the key we can use the `cache` (if present) and do not net to iterate over the trie nodes again.


cheme:
Is this the test ? https://github.com/paritytech/trie/blob/0b9bede27b33da6395e83ed13c523ed2ae913fc6/trie-db/src/lookup.rs#L441 None variant should work fine with it.

cheme:
oh no, forget about it :)

lexnv (approved):

cheme (approved):

Issue #164, opened 2022-08-27
Implement `DoubleEndedIterator` for `TrieDBNodeIterator` and `TrieDBIterator`


seunlanlege:
This would allow users to walk the trie backwards from a specified prefix. Rather than having to load all trie items into memory and perform this operation in a `Vec`.

snowmead:
Are there any advancements on this? This would be ideal for our use case.

PR #208, opened 2023-12-22
Implements `DoubleEndedIterator` for trie iterators

snowmead:
Closes https://github.com/paritytech/trie/issues/164

Enables bidirectional iteration by implementing the `DoubleEndedIterator` for all iterator types.

New double ended iterator structs:

- `TrieDBDoubleEndedIterator`
- `TrieDBNodeDoubleEndedIterator`
- `TrieDBKeyDoubleEndedIterator`
- `FatDBDoubleEndedIterator`

Internal iteration methods now accept a `fwd` bool flag to conditionally traverse a trie backwards or forwards (non exhaustive list).

- `TrieDBRawIterator::next_raw_item`
- `TrieDBRawIterator::seek`
- `Crumb::increment` (renamed to `Crumb::step` to better convey its behaviour)

Modifies `next_raw_item` to return middle node (i.e. a node containing child nodes) in reverse order based on iteration direction. Middle node is returned last when iterating backwards, mirroring forward iteration behaviour.

Adds new `AftExiting` status to pop crumb after returning it in `Exiting` status for backward iteration.

Library users can create one of these iterators from a `TrieDB` instance by calling the public `TrieDB::into_*` methods.

```rust
let trie = TrieDBBuilder::<T>::new(&memdb, &root).build();
// instantiates `TrieDBDoubleEndedIterator`
let mut iter = trie.into_double_ended_iter().unwrap();

// forward iteration
iter.next().unwrap().unwrap());
// backward iteration
iter.next_back()
```

### DIFF
--- a/trie-db/src/fatdb.rs
+++ b/trie-db/src/fatdb.rs
@@ -18,7 +18,7 @@ use super::{
 };
 use hash_db::{HashDBRef, Hasher};
 
-use crate::{rstd::boxed::Box, MerkleValue, TrieDBBuilder};
+use crate::{rstd::boxed::Box, triedb::TrieDBDoubleEndedIterator, MerkleValue, TrieDBBuilder};
 
 /// A `Trie` implementation which hashes keys and uses a generic `HashDB` backing database.
 /// Additionaly it stores inserted hash-key mappings for later retrieval.
@@ -109,6 +109,22 @@ where
 	trie: &'db TrieDB<'db, 'cache, L>,
 }
 
+/// Double ended iterator over inserted pairs of key values.
+pub struct FatDBDoubleEndedIterator<'db, 'cache, L>
+where
+	L: TrieLayout,
+{
+	trie_iterator: TrieDBDoubleEndedIterator<'db, 'cache, L>,
+	trie: &'db TrieDB<'db, 'cache, L>,
+}
+
+impl<'a, 'cache, L: TrieLayout> FatDBDoubleEndedIterator<'a, 'cache, L> {
+	/// Create a new double ended iterator.
+	pub fn new(db: &'a TrieDB<'a, 'cache, L>) -> Result<Self, TrieHash<L>, CError<L>> {
+		Ok(Self { trie_iterator: TrieDBDoubleEndedIterator::new(db)?, trie: db })
+	}
+}
+
 impl<'db, 'cache, L> FatDBIterator<'db, 'cache, L>
 where
 	L: TrieLayout,
@@ -148,7 +164,36 @@ where
 	}
 }
 
-impl<'db, 'cache, L> DoubleEndedIterator for FatDBIterator<'db, 'cache, L>
+impl<'db, 'cache, L> TrieIterator<L> for FatDBDoubleEndedIterator<'db, 'cache, L>
+where
+	L: TrieLayout,
+{
+	fn seek(&mut self, key: &[u8]) -> Result<(), TrieHash<L>, CError<L>> {
+		let hashed_key = L::Hash::hash(key);
+		self.trie_iterator.seek(hashed_key.as_ref())
+	}
+}
+
+impl<'db, 'cache, L> Iterator for FatDBDoubleEndedIterator<'db, 'cache, L>
+where
+	L: TrieLayout,
+{
+	type Item = TrieItem<TrieHash<L>, CError<L>>;
+
+	fn next(&mut self) -> Option<Self::Item> {
+		self.trie_iterator.next().map(|res| {
+			res.map(|(hash, value)| {
+				let aux_hash = L::Hash::hash(&hash);
+				(
+					self.trie.db().get(&aux_hash, Default::default()).expect("Missing fatdb hash"),
+					value,
+				)
+			})
+		})
+	}
+}
+
+impl<'db, 'cache, L> DoubleEndedIterator for FatDBDoubleEndedIterator<'db, 'cache, L>
 where
 	L: TrieLayout,
 {
--- a/trie-db/src/fatdb.rs
+++ b/trie-db/src/fatdb.rs
@@ -148,6 +148,23 @@ where
 	}
 }
 
+impl<'db, 'cache, L> DoubleEndedIterator for FatDBIterator<'db, 'cache, L>
+where
+	L: TrieLayout,
+{
+	fn next_back(&mut self) -> Option<Self::Item> {
+		self.trie_iterator.next_back().map(|res| {
+			res.map(|(hash, value)| {
+				let aux_hash = L::Hash::hash(&hash);
+				(
+					self.trie.db().get(&aux_hash, Default::default()).expect("Missing fatdb hash"),
+					value,
+				)
+			})
+		})
+	}
+}
+
 /// Iterator over inserted keys.
 pub struct FatDBKeyIterator<'db, 'cache, L>
 where
@@ -185,6 +202,20 @@ where
 
 	fn next(&mut self) -> Option<Self::Item> {
 		self.trie_iterator.next().map(|res| {
+			res.map(|hash| {
+				let aux_hash = L::Hash::hash(&hash);
+				self.trie.db().get(&aux_hash, Default::default()).expect("Missing fatdb hash")
+			})
+		})
+	}
+}
+
+impl<'db, 'cache, L> DoubleEndedIterator for FatDBKeyIterator<'db, 'cache, L>
+where
+	L: TrieLayout,
+{
+	fn next_back(&mut self) -> Option<Self::Item> {
+		self.trie_iterator.next_back().map(|res| {
 			res.map(|hash| {
 				let aux_hash = L::Hash::hash(&hash);
 				self.trie.db().get(&aux_hash, Default::default()).expect("Missing fatdb hash")
--- a/trie-db/src/fatdb.rs
+++ b/trie-db/src/fatdb.rs
@@ -254,17 +254,3 @@ where
 		})
 	}
 }
-
-impl<'db, 'cache, L> DoubleEndedIterator for FatDBKeyIterator<'db, 'cache, L>
-where
-	L: TrieLayout,
-{
-	fn next_back(&mut self) -> Option<Self::Item> {
-		self.trie_iterator.next_back().map(|res| {
-			res.map(|hash| {
-				let aux_hash = L::Hash::hash(&hash);
-				self.trie.db().get(&aux_hash, Default::default()).expect("Missing fatdb hash")
-			})
-		})
-	}
-}
--- a/trie-db/src/iterator.rs
+++ b/trie-db/src/iterator.rs
@@ -704,7 +704,8 @@ impl<L: TrieLayout> TrieDoubleEndedIterator<L> for TrieDBNodeDoubleEndedIterator
 
 impl<'a, 'cache, L: TrieLayout> TrieIterator<L> for TrieDBNodeDoubleEndedIterator<'a, 'cache, L> {
 	fn seek(&mut self, key: &[u8]) -> Result<(), TrieHash<L>, CError<L>> {
-		self.raw_iter.seek(self.db, key).map(|_| ())
+		self.raw_iter.seek(self.db, key).map(|_| ())?;
+		self.back_raw_iter.seek(self.db, key).map(|_| ())
 	}
 }
 
--- a/trie-db/src/iterator.rs
+++ b/trie-db/src/iterator.rs
@@ -78,24 +78,32 @@ impl<H: Hasher> Crumb<H> {
 pub struct TrieDBRawIterator<L: TrieLayout> {
 	/// Forward trail of nodes to visit.
 	trail: Vec<Crumb<L::Hash>>,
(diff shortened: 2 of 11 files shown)


trie-db/src/iterator.rs
```diff
@@ -56,31 +56,64 @@ impl<H: Hasher> Crumb<H> {
...
 	}
+
+	/// Move on to next status in the node's sequence in backward direction.
+	fn decrement(&mut self) {
+		self.status = match (self.status, self.node.node_plan()) {
+			(Status::Entering, NodePlan::Extension { .. }) => Status::At,
+			(Status::Entering, NodePlan::Branch { .. }) |
+			(Status::Entering, NodePlan::NibbledBranch { .. }) => Status::At,
+			(Status::At, NodePlan::Branch { .. }) |
+			(Status::At, NodePlan::NibbledBranch { .. }) => Status::AtChild(15),
```
cheme, line 67:
```suggestion
			(Status::At, NodePlan::NibbledBranch { .. }) => Status::AtChild(nibble_ops::NIBBLE_LENGTH - 1),
```

trie-db/src/iterator.rs
```diff
@@ -56,31 +56,64 @@ impl<H: Hasher> Crumb<H> {
...
 /// Iterator for going through all nodes in the trie in pre-order traversal order.
 pub struct TrieDBRawIterator<L: TrieLayout> {
+	/// Forward trail of nodes to visit.
 	trail: Vec<Crumb<L::Hash>>,
+	/// Forward iteration key nibbles of the current node.
 	key_nibbles: NibbleVec,
+	/// Backward trail of nodes to visit.
+	back_trail: Vec<Crumb<L::Hash>>,
+	/// Backward iteration key nibbles of the current node.
+	back_key_nibbles: NibbleVec,
```
cheme, line 86: Did you try simply using `trail` and `key_nibbles` for backward iteration? Not duplicating it would be good as it would make switching between next and previous call simple.
Also may remove the need to implement clone.

trie-db/src/iterator.rs
```diff
@@ -447,6 +489,104 @@ impl<L: TrieLayout> TrieDBRawIterator<L> {
 		}
 	}
 
+	fn next_back_raw_item(
```
cheme, line 492: if using same trail and crumb then we can avoid this duplicated method (just pass direction in parameter to `next_raw_item` function and use either increment or decrement depending on it. 
Actually we also could not have `increment` and `decrement` function but just `increment` with a parameter `direction`: would avoid lot of conditional here.

trie-db/src/iterator.rs
```diff
@@ -496,6 +636,52 @@ impl<L: TrieLayout> TrieDBRawIterator<L> {
 		None
 	}
 
+	pub fn next_back_item(&mut self, db: &TrieDB<L>) -> Option<TrieItem<TrieHash<L>, CError<L>>> {
```
cheme, line 639: Should be possible to factor code with `next_item` (eg have a internal non public function that take a direction as parameter).

cheme (changes requested):
Thanks for the PR.

trie-db/src/iterator.rs
snowmead: How would you track both cursors iterating forward and backwards? 

The `next_raw_item` takes in `&mut self` and passing a direction would not be enough to simutaneously iterate in both directions. Are you implying that we should add a `back_raw_iter` to every iterator struct like this?

```
/// Iterator for going through all nodes in the trie in pre-order traversal order.
pub struct TrieDBNodeIterator<'a, 'cache, L: TrieLayout> {
	db: &'a TrieDB<'a, 'cache, L>,
	raw_iter: TrieDBRawIterator<L>,
	back_raw_iter: TrieDBRawIterator<L>,
}
```

Or you think there should be a single cursor going backwards and forwards? (I think this would not be ideal, since you would need to iterate much more if you wanted to fetch 2 nodes simutaneously).

cheme: > Or you think there should be a single cursor going backwards and forwards? 

Yes, double end iterator should be able to do both.

> (I think this would not be ideal, since you would need to iterate much more if you wanted to fetch 2 nodes simutaneously).

If you need two nodes, then you would need two instance of iterator or am I missing something?

snowmead: > If you need two nodes, then you would need two instance of iterator or am I missing something?

Ok I see, so you think we should have a single cursor. I did not go for this initially since I was basing my implementation on the [`DoubleEndedIterators`](https://doc.rust-lang.org/std/iter/trait.DoubleEndedIterator.html#examples) documentation which demonstrates 2 cursors. 

In that case, I am for using a single cursor cause it would simplify the implementation.

cheme: Oh yes you are right, the actual trait is working on two pointers.
For trie crate it means doing what you propose:
```
// Iterator for going through all nodes in the trie in pre-order traversal order.
pub struct TrieDBNodeIterator<'a, 'cache, L: TrieLayout> {
	db: &'a TrieDB<'a, 'cache, L>,
	raw_iter: TrieDBRawIterator<L>,
	back_raw_iter: TrieDBRawIterator<L>,
}
```
would be better to be a different iterator (eg DoubleTrieDBNodeIterator), to keep the smaller struct when doubleended is not needed (keeping `next_back_item` public for it make sense to me, so we can still use the standard iterator in both direction (the use case of DoubleTrieDBNodeIterator is a bit off what I would expect).

snowmead: > (keeping `next_back_item` public for it make sense to me, so we can still use the standard iterator in both direction (the use case of DoubleTrieDBNodeIterator is a bit off what I would expect)

Not sure what you mean by this. Do you mean that you would still be able to iterate with a single cursor if using the raw iterator directly?

cheme: keeping TrieDBNodeIterator as is and having a new struct (as you describe) :
```
pub struct TrieDBNodeDoubleEndedIterator<'a, 'cache, L: TrieLayout> {
	db: &'a TrieDB<'a, 'cache, L>,
	raw_iter: TrieDBRawIterator<L>,
	back_raw_iter: TrieDBRawIterator<L>,
}
```
and a new function to get it (eg `fn double_ended_iter(self: &TrieDB) -> TrieDBNodeDoubleEndedIterator`)

trie-db/src/iterator.rs
snowmead: I was thinking of leaving `next_back_item` to avoid a breaking change in regards to the `next_item` api, since it's public function.

trie-db/src/iterator.rs
snowmead: I don't think we should allow iterating backwards with any of the standard iterators. To iterate backwards, they would have to use `TrieDBNodeDoubleEndedIterator`. 

The api is a little confusing having the possibility to iterate backwards with a single cursor for a regular iterator. There should be an explicit distinction between these two types.

Also iterating backwards with a single cursor can result in ending the iteration if it reaches the back of the trie. To me this is even more confusing, since the original implementation relies on reaching the end of the trie (eg the last node in the trie) to end. It also adds more complexity.

cheme: The change for iterating backward looks simple enough to me ( https://github.com/snowmead/trie/compare/master...cheme:trie:208 )
But I agree api gets a bit confusing when it comes to have a specific doubleendediterator struct.
But looking at my changes, it seems easy enough to just not use the trait (see change in the test).
I mean in our case the double ended nature is not part of our iterator and requires doubling the struct size.

> Also iterating backwards with a single cursor can result in ending the iteration if it reaches the back of the trie. 

Yes that s a good point, not sure how problematic it can be though.
Can be fix by always keeping root node or prefix node in the crumb list, but I am not sure the change is worth it.

Maybe we just want to expose a backward iterator (IIRC some std behavior for `.rev`  did surprise me a bit in the past, but as long as it is call before first iteration it is fairly ok).

snowmead:
@cheme 
> The modification for iterating backward seems straightforward to me.

However, I believe adding this behaviour in a standard iterator is unintuitive for library users.

>> Additionally, iterating backward with a single cursor might prematurely end the iteration when it reaches the back of the trie.

> That's a valid concern. It could potentially be addressed by always including the root node or prefix node in the crumb list. Yet, I am uncertain whether the change justifies the effort.

IMHO I don't think it does. I believe having separate structs for explicit iteration in both directions enhances code clarity. It also provides a better understanding of the behavior of standard iterator structs compared to double ended iterator structs.

I've implemented modifications based on our discussion:

- Renamed `increment` to [`step`](https://github.com/paritytech/trie/pull/208/files#diff-044be399afe1276a312ce355e3502e58b5d3a7af1c11c57409e2b2c4759aea95R45).
- A specific double ended iterator struct can be created from `TrieDB`. I'm unsure if a more elegant or generic solution is desired.
- Deduplicated all internal iteration functions and added the `fwd` bool parameter.
- Created double ended iterator structs for each type of iterator (e.g., node, key, etc.).

trie-db/src/iterator.rs
snowmead: Answered outside of this thread, since it impacts more then the initial comment. https://github.com/paritytech/trie/pull/208#issuecomment-1871442289

trie-db/src/fatdb.rs
```diff
@@ -109,6 +109,22 @@ where
 	trie: &'db TrieDB<'db, 'cache, L>,
 }
 
+/// Double ended iterator over inserted pairs of key values.
+pub struct FatDBDoubleEndedIterator<'db, 'cache, L>
```
cheme, line 113: I should have mention earlier, but fatdb iterator and fatdb generally is a bit dead code (old ethereum thing).
So maybe it is better not to have it (I know that it was remove in some substrate specific trie branches), but no harm having it.

trie-db/src/iterator.rs
```diff
@@ -450,8 +473,15 @@ impl<L: TrieLayout> TrieDBRawIterator<L> {
...
 	/// Must be called with the same `db` as when the iterator was created.
-	pub fn next_item(&mut self, db: &TrieDB<L>) -> Option<TrieItem<TrieHash<L>, CError<L>>> {
-		while let Some(raw_item) = self.next_raw_item(db) {
+	///
+	/// Specify `fwd` to indicate the direction of the iteration (`true` for forward, `false` for
+	/// backward).
+	pub fn next_item(
+		&mut self,
+		db: &TrieDB<L>,
+		fwd: bool,
```
cheme, line 482: So this change a "not so public" api (raw db iterator should not be use everywhere), so should be remember this require correct versioning changes when publishing.

cheme: (could also make a new function `prev_item`, but in this case an internal function with the parameter seems still needed to avoid code redundancy).

cheme (approved):
Thanks for the  changes, `into_double_ended_iter` api looks good to me.

trie-db/src/iterator.rs
snowmead: > So this change a "not so public" api (raw db iterator should not be use everywhere), so should be remember this require correct versioning changes when publishing.

You mean it's public but not many people should be using it? We would still need to bump the version to the correct one to convey it's breaking the api. 

> (could also make a new function `prev_item`, but in this case an internal function with the parameter seems still needed to avoid code redundancy).

I initially did this, but I found it much cleaner to just have `next_item`. I could also extract the core logic of the function into a reusable one so that there is no redundant code.

trie-db/src/fatdb.rs
snowmead: I think I'll just leave it if it isn't a big deal. If you insist, I will remove it.

cheme:
Small note (not something that needs change right now), I think that the raw node iter can return on a status At, then if iterating in different direction we may not have the right next node : eg entering a branch backward then state is At then calling iterator forward and then state become AtChild(0) when we should have exit.
This is not an issue right now as we don't expose different iteration direction, but at some point I will probably switch to state At(direction: bool) to be able to support this case.

snowmead:
> Small note (not something that needs change right now), I think that the raw node iter can return on a status At, then if iterating in different direction we may not have the right next node : eg entering a branch backward then state is At then calling iterator forward and then state become AtChild(0) when we should have exit.

I am not sure to entirely understand the scenario. It sounds like this describes a case where you call the raw iterator directly without any struct impl and call forwards and backwards simutaneously. Basically having a single cursor moving in both directions.

snowmead:
@cheme I am adding a test for seeking and iterating backwards, but I am not sure if the behaviour makes sense. 

It panics when I seek for `"0125"` prefix which should yield at least some node when executing `next_back()`, but it yields `None` instead. Not sure what might be causing this behaviour. I wouldn't think the `seek` method needs to be modified since it gives you the node at that prefix or the one after. Afterwards, you can iterate backwards as you would normally from a node, but maybe I am wrong.

https://github.com/paritytech/trie/pull/208/files#diff-022f95ed46c85f83274f42b2d1b9d26483ba910fa31430332a4fc5d948a02573R237

arkpar (approved):

PR #210, opened 2024-01-22
new trie

cheme:
Opening this PR to prepare the branch to start the new repo (not for merge here).

Based on https://github.com/paritytech/trie/tree/arkpar/new-trie

Draft for now.

Substrate branch: https://github.com/cheme/polkadot-sdk/tree/cheme/new-trie also based on https://github.com/paritytech/polkadot-sdk/tree/arkpar/new-trie


trie-db/src/lib.rs
```diff
@@ -530,21 +363,29 @@ pub trait TrieLayout {
 	type Hash: Hasher;
 	/// Codec to use (needs to match hasher and nibble ops).
 	type Codec: NodeCodec<HashOut = <Self::Hash as Hasher>::Out>;
+	type Location: Location;
```
cheme, line 366: `const USE_EXTENSION` and all extension related code looks easy enough to remove to me, but can be the first pr of the new code base.

trie-db/src/lookup.rs
```diff
@@ -22,20 +22,22 @@ use crate::{
...
 pub struct Lookup<'a, 'cache, L: TrieLayout, Q: Query<L::Hash>> {
 	/// database to query from.
-	pub db: &'a dyn HashDBRef<L::Hash, DBValue>,
+	pub db: &'a dyn HashDB<L::Hash, DBValue, L::Location>,
 	/// Query object to record nodes and transform data.
 	pub query: Q,
 	/// Hash to start at
 	pub hash: TrieHash<L>,
+	/// Optionally location to start at.
+	pub location: L::Location,
```
cheme, line 36: From arkpar/new-trie I did add location to lookup which remove the need to pass it in parameter in lookup api.

trie-db/src/node.rs
```diff
@@ -577,29 +638,67 @@ impl NodePlan {
...
+						count += 1;
+					}
+				}
+				count
+			},
+			_ => 0,
+		}
+	}
+
+	pub fn attached_change_set_location<L: Copy + Default>(&self, locations: &[L]) -> Option<L> {
```
cheme, line 664: I did extract this to be able to still work on nodeplan up to a certain extent (changes from arkpar/new-trie in iterator.rs when decoding nodes), but I think location should just be included into nodeplan

hash-db/src/lib.rs
```diff
@@ -73,144 +78,19 @@ pub trait Hasher: Sync + Send {
 	fn hash(x: &[u8]) -> Self::Out;
 }
 
-/// Trait modelling a plain datastore whose key is a fixed type.
```
cheme, line 76: make sense to remove all these traits :+1:

trie-db/src/triedbmut.rs
```diff
@@ -634,65 +652,78 @@ impl<'a, L: TrieLayout> Index<&'a StorageHandle> for NodeStorage<L> {
...
 	) -> Self {
-		Self { db, root, cache: None, recorder: None }
+		Self { db, root, cache: None, recorder: None, root_location: Default::default() }
+	}
+
+	/// Same as `from_existing` but force a db location to access root.
+	/// Note root in parameter is not checked.
+	pub fn from_existing_with_db_location(
+		db: &'db dyn HashDB<L::Hash, DBValue, L::Location>,
+		root: TrieHash<L>,
```
cheme, line 692: (change from arkpar/new-trie)
Here, root is not really usefull, since we use location instead. Maybe something to do, either force a check or removing this parameter. (also true for triedbmut)

trie-db/src/triedbmut.rs
```diff
@@ -711,18 +742,144 @@ impl<'db, L: TrieLayout> TrieDBMutBuilder<'db, L> {
...
 }
 
+#[derive(Debug)]
+pub struct NewChangesetNode<H, DL> {
+	pub hash: H,
+	pub prefix: OwnedPrefix,
+	pub data: Vec<u8>,
+	pub children: Vec<ChangesetNodeRef<H, DL>>,
+	// Storing the key and removed nodes (only needed for old trie).
+	pub key_childset: Option<(Vec<u8>, Vec<(H, OwnedPrefix)>)>,
```
cheme, line 758: change from arkpar/new-trie for substrate child trie usecase support.
Not that this is not used when using locations, but I don't think it is worth making conditional.
On the other hand not feeding the removed nodes and debug asserting it is empty when using location may be a good idea.

trie-db/src/triedbmut.rs
```diff
@@ -711,18 +742,144 @@ impl<'db, L: TrieLayout> TrieDBMutBuilder<'db, L> {
...
+			root: ChangesetNodeRef::Existing(ExistingChangesetNode {
+				hash: root,
+				prefix: (BackingByteVec::new(), None),
+				location: Default::default(),
+			}),
+			removed: Default::default(),
+		}
+	}
+
+	pub fn to_insert_in_other_trie(mut self, root_at: Vec<u8>) -> Box<ChangesetNodeRef<H, DL>> {
```
cheme, line 862: changes from arkpar/new-trie. Probably not the best naming.

trie-db/src/triedbmut.rs
```diff
@@ -1804,152 +2034,120 @@ where
...
-			} else {
-				None
-			};
-
-			if let Some(value) = value {
-				cache.cache_value_for_key(full_key, (value, hash).into())
-			}
-		}
-	}
+	fn cache_value(&mut self, _full_key: &[u8], _value: impl Into<Bytes>, _hash: TrieHash<L>) {}
```
cheme, line 2150: Generally arkpar/new-trie seems to have disable/remove some cache usage with triedbmut, is it something that should be restore later?

trie-db/src/triedbmut.rs
```diff
@@ -2014,53 +2240,65 @@ where
...
-	fn insert(
+	pub fn insert(
 		&mut self,
 		key: &[u8],
 		value: &[u8],
 	) -> Result<Option<Value<L>>, TrieHash<L>, CError<L>> {
+		self.insert_with_child_changes(key, value, None)
+	}
+
+	pub fn insert_with_child_changes(
```
cheme, line 2287: changes from arkpar/new-trie

trie-db/src/triedbmut.rs
```diff
@@ -2082,15 +2325,33 @@ where
 		Ok(old_val)
 	}
 
-	fn remove(&mut self, key: &[u8]) -> Result<Option<Value<L>>, TrieHash<L>, CError<L>> {
+	pub fn remove(&mut self, key: &[u8]) -> Result<Option<Value<L>>, TrieHash<L>, CError<L>> {
+		self.remove_with_child_changes(key, None)
+	}
+
+	pub fn remove_with_child_changes(
```
cheme, line 2332: change from arkpar/new-trie

trie-db/src/triedbmut.rs
```diff
@@ -2099,24 +2360,16 @@ where
...
 }
 
-impl<'a, L> Drop for TrieDBMut<'a, L>
-where
-	L: TrieLayout,
-{
-	fn drop(&mut self) {
-		self.commit();
-	}
-}
```
cheme, line 2118: agree with the removal, this action should be implicit I think. TODO Maybe worth putting in release note.

trie-db/test/src/triedbmut.rs
```diff
@@ -881,12 +848,205 @@ fn test_two_assets_memory_db_inner_2<T: TrieLayout>() {
...
 	let key3 = [1u8, 3];
 	let data3 = [3u8; 2];
 	state.insert(key3.as_ref(), &data3).unwrap();
 	assert_eq!(state.get(key1.as_ref()).unwrap().unwrap(), data1);
 	assert_eq!(state.get(key2.as_ref()).unwrap().unwrap(), data2);
 	assert_eq!(state.get(key3.as_ref()).unwrap().unwrap(), data3);
 }
+
+test_layouts!(child_trie, child_trie_internal);
+fn child_trie_internal<T: TrieLayout, DB: TestDB<T>>() {
```
cheme, line 860: changes from arkpar/new-trie important to review as gives a good view on usage for substrate (see also https://github.com/cheme/polkadot-sdk/tree/cheme/new-trie where I find the use of iterator a bit awkward when calculating root but a given struct encapsulating it may make things ok).

cheme (commented):
@arkpar I did add review comments, some of them propose changes, some are about your initial branch and some points to my changes.
Code looks good to me (cannot approve because I opened the PR).
But I also added my changes for child tree support to the branch , some naming are terribly wrong, and can be reviewed, I tried to add comments where the big changes are.
Also wondering about putting location always attached to trie hash, I mean every where, would probably change the code a bit/lot, but may make things simpler? (thinking twice may be a bit confusing)

Also attached changeset in triedbmut can be overwritten easily: not in substrate case as the keys for child trie are not writable. Was thinking on putting a drop implementation in the attached changeset that ensure the changeset was properly consumed (same for standard changeset unless drop properly as dropping a changeset in code is probably not a good idea (useless cost of creating it)).

trie-db/src/triedbmut.rs
```diff
@@ -1365,24 +1583,28 @@ where
 		node: Node<L>,
 		key: &mut NibbleFullKey,
 		old_val: &mut Option<Value<L>>,
+		child_changes: Option<ChildChangeset<L>>,
```
arkpar, line 1586: `tree_ref` here and everywhere

trie-db/src/triedbmut.rs
arkpar: Basically we can't cache a node before we know its location info. Otherwise if we add a node with no location info, it may be later retrieved from the cache and the database will fail to locate children.

trie-db/src/triedbmut.rs
arkpar: `insert_with_tree_ref`

trie-db/src/triedbmut.rs
arkpar: `remove_with_tree_ref`

I guess eventuall we won't need this? The point of tree references is that the database handles dereferencing them automatically.

trie-db/src/triedbmut.rs
arkpar: `self.db` is read-only anyway, so drop can't really do anything.
There's a lot of changes in the API, hence the new crate.

trie-db/src/triedbmut.rs
arkpar, line 1: I would not use the term "child" for subtries or tree references. We already use "children" all over this crate to refer to children of branch nodes. Using the same term for tree references with no extra context is confusing.

trie-db/test/src/triedbmut.rs
```diff
@@ -881,12 +848,205 @@ fn test_two_assets_memory_db_inner_2<T: TrieLayout>() {
...
+			&keyspaced_memdb
+		};
+		let mut child_trie = TrieDBMutBuilder::<T>::from_existing_with_db_location(
+			child_memdb,
+			a_child_trie_root,
+			child_trie_location.unwrap_or_default(),
+		)
+		.build();
+		child_trie.remove(a_child_trie.data.iter().next().unwrap().0).unwrap();
+		child_trie.insert(b"make_sur_it_changes", b"value").unwrap();
```
arkpar, line 964: make_sure

trie-db/test/src/triedbmut.rs
arkpar: Loogs good, apart from the naming.

arkpar (changes requested):

trie-db/src/triedbmut.rs
cheme: True, might need to pass this `root_at` as param on commit (as Option<Vec<u8>>), or a different commit function when needed.
(root_at is only needed for rocksdb HashDB prefixing).

hash-db/src/lib.rs
```diff
@@ -73,144 +78,19 @@ pub trait Hasher: Sync + Send {
...
-impl<'a, K, V> PlainDBRef<K, V> for &'a dyn PlainDB<K, V> {
-	fn get(&self, key: &K) -> Option<V> {
-		PlainDB::get(*self, key)
-	}
-	fn contains(&self, key: &K) -> bool {
-		PlainDB::contains(*self, key)
-	}
-}
+/// Trait modelling datastore keyed by a hash defined by the `Hasher` and optional location tag.
+pub trait HashDB<H: Hasher, T, L>: Send + Sync {
```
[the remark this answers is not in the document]
cheme: Should we also change traits to return errors ? I mean if we return error I would go for NodeDB, if not NodeStore.

(also wondering about renaming crate here)

cheme:
> Basically we can't cache a node before we know its location info. Otherwise if we add a node with no location info, it may be later retrieved from the cache and the database will fail to locate children.

I see, probably some things to do in the future, but seems better to start this way.

hash-db/src/lib.rs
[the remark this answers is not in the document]
arkpar: > Should we also change traits to return errors ?

Probably, but in a separate PR.

cheme:
> Might be a good idea to rename this to NodeDB or NodeStore at this point

Question, if we do a new trie crate, suposedly with a new name.
We would need a new hashdb crate (so nodedb), a new memory-db crate (no idea for the name here) and also a new trie-bench.
Trie_root would not really need a new name.

Or just keeping old name and using new repo would be the idea (would make previous crate publishing difficult)?

arkpar:
The new repo is supposed to be this:
https://github.com/paritytech/subtrie

Also, I guess we can put everythung into a single crate for simplicity.

cheme:
In https://github.com/paritytech/trie/pull/210/commits/28a5476ac9b30faff2eb650e1c21740ca9d0d0a9 I did merge the memory db mem tree db and hash db crate in trie one.
Looks good I think, except I chose to keep using `Hasher` trait from old hash_db (but I deleted the dep in the project as it should be very stable and only this trait would be used).
When trying to use `Hasher` in trie-db crate I end up needing keccack hasher to depend on triedb, but is use in memorydb test so I moved memorydb test to trie-db-test, but then I all private field access are problematic.

Another way to do it would be to move keccak hasher  and trie root in triedb, maybe it is a good idea actually, may do tomorrow (only bad point is that we duplicate code here, I mean old crate and new crate, so not totally sure).

cheme:
In https://github.com/paritytech/trie/pull/210/commits/554187bd82e8b3db5ae5463395d74e8720fb65f3 I did remove lot of crates, only keeping h256.
I could have kept trie-root, but then old crates like hash-db would have been added to dependencies in substrate.
keccak hasher is a bit awkward, could use a real test hasher (with no dependency) in place but tiny keccak should be behind a feature.

cheme:
Also did flatten the crate hieriarchy since there is not many left.

PR #214, opened 2024-04-09
HashDb trait should not imply Send and Sync

ufoscout:
The HashDB trait mandates its implementors to be Send and Sync. However, since removing this constraint doesn't affect the remaining code, I'm curious if there's a particular reason for retaining it. 
I'm particularly interested in this because I need to implement the HashDB trait for a structure within a single-threaded wasm runtime that isn't Send or Sync.


cheme:
Yes I think there is no real need to be send sync at trait level.

cheme (approved):

PR #216, opened 2025-02-10
Improve TrieCache size by reducing size_of NodeOwned

alexggh:
NodeOwned uses static arrays to represent the children of a `NodeOwned::Branch` and `NodeOwned::NibbledBranch`, because the enum's  size is  the size of largest variant it ends up occupying 752 bytes. 

The substrate ShareTrieCache ends up keeping  a hash map with elements of NodeOwned, so it's size quickly balloons, although most of the children seem to be empty. 

The trie cache size benefits greatly from using a dynamic array instead of static one: https://github.com/paritytech/polkadot-sdk/issues/6131#issuecomment-2621222929, using westend-asset-hub, we can see a ~4 time reduction in memory size for the TrieCache.

Fixes: https://github.com/paritytech/polkadot-sdk/issues/7386

### DIFF
--- a/trie-db/src/node.rs
+++ b/trie-db/src/node.rs
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use core::iter;
+
 use crate::{
 	nibble::{self, nibble_ops, NibbleSlice, NibbleVec},
 	node_codec::NodeCodec,
@@ -210,34 +212,57 @@ impl Node<'_> {
 			Self::Extension(n, h) =>
 				Ok(NodeOwned::Extension((*n).into(), h.to_owned_handle::<L>()?)),
 			Self::Branch(childs, data) => {
-				let mut childs_owned = [(); nibble_ops::NIBBLE_LENGTH].map(|_| None);
+				let last_existing_child_index = childs
+					.iter()
+					.enumerate()
+					.flat_map(|(index, value)| value.map(|_v| index))
+					.rev()
+					.next()
+					.unwrap_or_default();
+				let mut childs_owned = Vec::with_capacity(last_existing_child_index + 1);
+
 				childs
 					.iter()
 					.enumerate()
 					.map(|(i, c)| {
-						childs_owned[i] =
-							c.as_ref().map(|c| c.to_owned_handle::<L>()).transpose()?;
+						if i <= last_existing_child_index {
+							childs_owned
+								.push(c.as_ref().map(|c| c.to_owned_handle::<L>()).transpose()?);
+						}
 						Ok(())
 					})
 					.collect::<Result<_, _, _>>()?;
 
-				Ok(NodeOwned::Branch(childs_owned, data.as_ref().map(|d| d.to_owned_value::<L>())))
+				Ok(NodeOwned::Branch(
+					ChildrenNodesOwned(childs_owned),
+					data.as_ref().map(|d| d.to_owned_value::<L>()),
+				))
 			},
 			Self::NibbledBranch(n, childs, data) => {
-				let mut childs_owned = [(); nibble_ops::NIBBLE_LENGTH].map(|_| None);
+				let last_existing_child_index = childs
+					.iter()
+					.enumerate()
+					.flat_map(|(index, value)| value.map(|_v| index))
+					.rev()
+					.next()
+					.unwrap_or_default();
+				let mut childs_owned = Vec::with_capacity(last_existing_child_index + 1);
+
 				childs
 					.iter()
 					.enumerate()
 					.map(|(i, c)| {
-						childs_owned[i] =
-							c.as_ref().map(|c| c.to_owned_handle::<L>()).transpose()?;
+						if i <= last_existing_child_index {
+							childs_owned
+								.push(c.as_ref().map(|c| c.to_owned_handle::<L>()).transpose()?);
+						}
 						Ok(())
 					})
 					.collect::<Result<_, _, _>>()?;
 
 				Ok(NodeOwned::NibbledBranch(
 					(*n).into(),
-					childs_owned,
+					ChildrenNodesOwned(childs_owned),
 					data.as_ref().map(|d| d.to_owned_value::<L>()),
 				))
 			},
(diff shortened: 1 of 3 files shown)


AndreiEres (approved):

trie-db/src/node.rs
```diff
@@ -210,34 +212,57 @@ impl Node<'_> {
...
+				))
 			},
 			Self::NibbledBranch(n, childs, data) => {
-				let mut childs_owned = [(); nibble_ops::NIBBLE_LENGTH].map(|_| None);
+				let last_existing_child_index = childs
+					.iter()
+					.enumerate()
+					.flat_map(|(index, value)| value.map(|_v| index))
+					.rev()
+					.next()
```
bkchr, line 247:
```suggestion
					.rposition(|v| v.is_some())
```

trie-db/src/node.rs
```diff
@@ -257,20 +282,46 @@ pub enum NodeOwned<H> {
...
+/// Before this struct was introduced, the children were stored in a fixed size array of length
+/// `nibble_ops::NIBBLE_LENGTH`, to make sure there aren't any direct accesses to the array that
+/// we've missed, wrap in its on struct, rather than directly use the `Vec`.
+#[derive(Eq, PartialEq, Clone)]
+#[cfg_attr(feature = "std", derive(Debug))]
+pub struct ChildrenNodesOwned<H>(Vec<Option<NodeHandleOwned<H>>>);
+
+impl<H> ChildrenNodesOwned<H> {
+	pub fn get(&self, index: usize) -> Option<&Option<NodeHandleOwned<H>>> {
+		self.0.get(index)
```
bkchr, line 313:
```suggestion
	pub fn get(&self, index: usize) -> Option<&NodeHandleOwned<H>> {
		self.0.get(index).map_or(None, |v| v.as_ref())
```

If a entry doesn't exist, it means it was `None`. Thus, we can simplify it.

bkchr: Some docs for the functions would also be nice.

trie-db/src/node.rs
```diff
@@ -257,20 +282,46 @@ pub enum NodeOwned<H> {
...
+	}
+
+	pub fn iter(&self) -> impl Iterator<Item = &Option<NodeHandleOwned<H>>> {
+		self.0.iter()
```
bkchr, line 317:
```suggestion
		self.0.iter().chain(std::iter::repeat(&None)).take(nibble_ops::NIBBLE_LENGTH)
```

This way it always returns the correct number of items.

trie-db/src/node.rs
```diff
@@ -289,16 +340,29 @@ where
...
+				C::branch_node_nibbled(
+					partial.right_iter(),
+					partial.len(),
+					children
+						.iter()
+						.chain(repeat)
+						.map(|child| child.as_ref().map(|c| c.as_child_reference::<C>())),
+					value.as_ref().map(|v| v.as_value()),
+				)
+			},
```
bkchr, line 365: Then we can revert these changes.

trie-db/src/node.rs
```diff
@@ -332,7 +396,7 @@ where
 								*index += 1;
 
 								// Ignore non-existing childs.
-								if let Some(ref child) = childs[*index - 1] {
+								if let Some(Some(ref child)) = childs.get(*index - 1) {
```
bkchr, line 399: Also these.

bkchr (commented):
Nice work :+1: 

I just want some improvements and then we almost need no changes.

alexggh:
> Nice work 👍
> 
> I just want some improvements and then we almost need no changes.

Done, thank you!

trie-db/src/node.rs
```diff
@@ -257,20 +284,52 @@ pub enum NodeOwned<H> {
...
+	/// The size of the iterator is always constant goes up to `nibble_ops::NIBBLE_LENGTH`, with
+	/// None representing the missing children.
+	pub fn iter(&self) -> impl Iterator<Item = &Option<NodeHandleOwned<H>>> {
+		self.0.iter().chain(iter::repeat(&None)).take(nibble_ops::NIBBLE_LENGTH)
+	}
+
+	/// Returns the number of children.
+	pub fn len(&self) -> usize {
+		self.0.len()
+	}
```
lexnv, line 330: nit: Might be good to have an `is_empty` for a complete API if we are exposing this to the user (https://rust-lang.github.io/rust-clippy/master/index.html#len_without_is_empty)

lexnv (approved):
Nicely done! 4x reduction is amazing 🚀

bkchr (approved):

trie-db/src/node.rs
```diff
@@ -257,20 +284,52 @@ pub enum NodeOwned<H> {
...
+/// None]` we will only store in the vector `[Some(_), None, Some(_)]`.
+///
+/// This is useful to save space in the TrieCache when the children are not fully populated.
+///
+/// Before this struct was introduced, the children were stored in a fixed size array of length
+/// `nibble_ops::NIBBLE_LENGTH`, to make sure there aren't any direct accesses to the array that
+/// we've missed, wrap it in its own struct, rather than directly use the `Vec`.
+#[derive(Eq, PartialEq, Clone)]
+#[cfg_attr(feature = "std", derive(Debug))]
+pub struct ChildrenNodesOwned<H>(Vec<Option<NodeHandleOwned<H>>>);
```
cheme, line 311: Not for this PR, but if the target is to strictly reduce memory size, then we should include the node bitmap in this struct too. Something like `ChindrenNodesOwned<>(Bitmap, Vec<NodeHandleOwned<H>>);` where get first check bit map and the vec only store present nodes (mapping node index should not be costy with bit operation over 2byte, but update would be rather costy : resizing Vec and rearranging data, but can be fine I guess, would need to be benchmarked).

trie-db/src/node.rs
```diff
@@ -257,20 +284,52 @@ pub enum NodeOwned<H> {
...
+	/// Returns an iterator over all children.
+	///
+	/// The size of the iterator is always constant goes up to `nibble_ops::NIBBLE_LENGTH`, with
+	/// None representing the missing children.
+	pub fn iter(&self) -> impl Iterator<Item = &Option<NodeHandleOwned<H>>> {
+		self.0.iter().chain(iter::repeat(&None)).take(nibble_ops::NIBBLE_LENGTH)
+	}
+
+	/// Returns the number of children.
+	pub fn len(&self) -> usize {
```
cheme, line 328: Could be renamed to `allocated_len`  to avoid confusion.

trie-db/src/node.rs
```diff
@@ -210,34 +212,59 @@ impl Node<'_> {
...
+								childs_owned.push(
+									c.as_ref().map(|c| c.to_owned_handle::<L>()).transpose()?,
+								);
+							}
+							Ok(())
+						})
+						.collect::<Result<_, _, _>>()?;
+					childs_owned
+				} else {
+					Vec::new()
```
cheme, line 262: Should specify capacity explicitely here (no real reason to specify it when there is a node and not here). (same in previous code).

cheme (approved):
Looks good to me.

trie-db/src/node.rs
alexggh: Will add it as a followup.

trie-db/src/node.rs
alexggh: You mean calling it Vec::with_capacity(0) ?

cheme: yes or something else if consider better, just to not depend on default behavior of new

PR #221, opened 2025-05-22
Replace BTreeMap usage with hashbrown hashmap

alexggh:
Discovered while profiling https://github.com/paritytech/polkadot-sdk/issues/6131#issuecomment-2891523233 with the benchmark https://github.com/paritytech/polkadot-sdk/pull/8069 that  when running in validation a big chunk of the time is spent inserting and retrieving data from the BTreeMap/BTreeSet.

This PR together with https://github.com/paritytech/polkadot-sdk/pull/8606 improve read costs with around ~40% and write costs with about ~20%

### DIFF
--- a/memory-db/src/lib.rs
+++ b/memory-db/src/lib.rs
@@ -30,7 +30,7 @@ use std::{
 };
 
 #[cfg(not(feature = "std"))]
-use alloc::collections::btree_map::{BTreeMap as Map, Entry};
+use hashbrown::{hash_map::Entry, HashMap as Map};
 
 #[cfg(not(feature = "std"))]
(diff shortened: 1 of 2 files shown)


memory-db/src/lib.rs
```diff
@@ -30,7 +30,7 @@ use std::{
 };
 
 #[cfg(not(feature = "std"))]
-use alloc::collections::btree_map::{BTreeMap as Map, Entry};
+use hashbrown::{hash_map::Entry, HashMap as Map};
```
bkchr, line 33: We can just use it on `std` and `no_std`. The `std` hashmap is actually hashbrown any way.

bkchr (approved):

skunert (approved):

AndreiEres (approved):

PR #224, opened 2025-07-08
memory-db: allow changing the default hasher

alexggh:
To be used by https://github.com/paritytech/polkadot-sdk/pull/9127, to replace the default hasher with a custom one.

### DIFF
--- a/memory-db/src/lib.rs
+++ b/memory-db/src/lib.rs
@@ -31,7 +31,10 @@ use std::{
 };
 
 #[cfg(not(feature = "std"))]
-use hashbrown::{hash_map::Entry, DefaultHashBuilder as RandomState, HashMap as Map};
+use hashbrown::{hash_map::Entry, HashMap as Map};
+
+#[cfg(not(feature = "std"))]
+use foldhash::quality::RandomState;
 
 #[cfg(not(feature = "std"))]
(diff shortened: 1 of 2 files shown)


skunert (approved):

bkchr:
@alexggh you should directly bump the crate version.

memory-db/src/lib.rs
```diff
@@ -238,23 +242,25 @@ pub fn legacy_prefixed_key<H: KeyHasher>(key: &H::Out, prefix: Prefix) -> Vec<u8
 	prefixed_key
 }
 
-impl<H, KF, T> Default for MemoryDB<H, KF, T>
+impl<H, KF, T, S> Default for MemoryDB<H, KF, T, S>
 where
 	H: KeyHasher,
 	T: for<'a> From<&'a [u8]>,
 	KF: KeyFunction<H>,
+	S: BuildHasher + Default,
```
bkchr, line 250:
```suggestion
impl<H, KF, T> Default for MemoryDB<H, KF, T, RandomState>
where
H: KeyHasher,
T: for<'a> From<&'a [u8]>,
KF: KeyFunction<H>,
```

This way you can just use `default()` without declaring the random state. Similar to `HashMap` we should then provide a function `with_random_state`

lexnv (approved):

alexggh:
> @alexggh you should directly bump the crate version.

Done.

PR #225, opened 2025-07-11
fixup! memory-db: allow changing the default hasher (#224)

alexggh:
Somehow, because cargo build is using the local reference of hashdb, cargo build did not complain about:

```
error[E0277]: `S` cannot be shared between threads safely
   --> src/lib.rs:604:3
    |
604 |         self
    |         ^^^^ `S` cannot be shared between threads safely
```

But cargo publish is using the `crates.io` version and in that case we get the above errors, not completely sure why the local  crate makes a difference, but I just fixed it.

Additionally, also set the `default-features` to true for foldhash, so that no_std build in polkadot-sdk works.

### DIFF
--- a/memory-db/Cargo.toml
+++ b/memory-db/Cargo.toml
@@ -10,7 +10,7 @@ edition = "2018"
 [dependencies]
 hash-db = { version = "0.16.0", path = "../hash-db", default-features = false }
 hashbrown = "0.15.3"
-foldhash = "0.1.5"
+foldhash = { version = "0.1.5", default-features = false }
 
 [dev-dependencies]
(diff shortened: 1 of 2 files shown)


lexnv (approved):

bkchr (approved):

michalkucharczyk:
Was this already released as memory-db `0.34.0` (likely with #224)? If yes could we have a tag please (could not found it)?

alexggh:
> Was this already released as memory-db `0.34.0` (likely with #224)? If yes could we have a tag please (could not found it)?

Sorry, forgot to push the tag it is `memory-db-v0.34.0`.

PR #226, opened 2025-09-23
`TrieDBMutBase` added: no commit on drop

michalkucharczyk:
This PR adds an option to prevent `TrieDBMut` from committing implicitly on drop.

This PR extends the `TrieDBMut` implementation to provide better control over when changes are committed to the database.

Changes:
- `commit_on_drop` flag added to `TrieDBMut`,
- Added `without_commit_on_drop`  method to `TrieDBMutBuilder` which configures the builder to create a _non-commiting_ `TrieDBMut` instance for cases where commit on drop is undesired.

Part of: https://github.com/paritytech/polkadot-sdk/issues/6020

### DIFF
--- a/trie-db/Cargo.toml
+++ b/trie-db/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "trie-db"
-version = "0.30.0"
+version = "0.31.0"
 authors = ["Parity Technologies <admin@parity.io>"]
 description = "Merkle-Patricia Trie generic over key hasher and node encoding"
 repository = "https://github.com/paritytech/trie"
--- a/trie-db/src/triedbmut.rs
+++ b/trie-db/src/triedbmut.rs
@@ -635,7 +635,7 @@ impl<'a, L: TrieLayout> Index<&'a StorageHandle> for NodeStorage<L> {
 	}
 }
 
-/// A builder for creating a [`TrieDBMut`].
+/// A builder for creating a [`TrieDBMut`] or [`TrieDBMutBase`].
 pub struct TrieDBMutBuilder<'db, L: TrieLayout> {
 	db: &'db mut dyn HashDB<L::Hash, DBValue>,
 	root: &'db mut TrieHash<L>,
@@ -738,8 +738,8 @@ impl<'db, L: TrieLayout> TrieDBMutBuilder<'db, L> {
 ///
 /// Querying the root of the trie will commit automatically.
 ///
-/// Dropping will not commit, refer to [`TrieDBMut`] for `Trie` implementation that performs
-/// `commit` on `drop`..
+/// Dropping the instance will not result in commit. Refer to [`TrieDBMut`] for `Trie`
+/// implementation that performs `commit` on `drop`.
 ///
 ///
 /// # Example
@@ -2186,9 +2186,16 @@ where
 	}
 }
 
-/// Type alias for `Trie` implementation.
+/// Type alias for `Trie` implementation using a generic `HashDB` backing database.
 ///
-/// Querying the root or dropping the trie will commit automatically.
+/// Can be used as a `TrieMut` trait object. `db()` can be used to get the backing database
+/// object.
+///
+/// Note that no changes are committed to the database until `commit` is called.
+///
+/// Querying the root or dropping the instance will commit automatically.
+///
+/// Refer to [`TrieDBMutBase`] for `Trie` implementation that does not perform `commit` on `drop`.
 pub type TrieDBMut<'a, L> = CommitOnDrop<'a, L>;
 
 /// combine two NodeKeys
--- a/trie-db/src/triedbmut.rs
+++ b/trie-db/src/triedbmut.rs
@@ -705,7 +705,7 @@ impl<'db, L: TrieLayout> TrieDBMutBuilder<'db, L> {
 	/// This is useful when you want fine-grained control over when changes are committed, or when
 	/// you want to avoid the performance cost of committing if you're just doing temporary
 	/// operations.
-	pub fn without_commit_on_drop(mut self) -> Self {
+	pub fn disable_commit_on_drop(mut self) -> Self {
 		self.commit_on_drop = false;
 		self
 	}
@@ -713,7 +713,7 @@ impl<'db, L: TrieLayout> TrieDBMutBuilder<'db, L> {
 	/// Build the [`TrieDBMut`].
 	///
 	/// By default, the returned trie will automatically commit changes when dropped. Use
-	/// [`without_commit_on_drop`](Self::without_commit_on_drop) to disable this behavior.
+	/// [`disable_commit_on_drop`](Self::disable_commit_on_drop) to disable this behavior.
 	pub fn build(self) -> TrieDBMut<'db, L> {
 		let root_handle = NodeHandle::Hash(*self.root);
 
@@ -737,9 +737,7 @@ impl<'db, L: TrieLayout> TrieDBMutBuilder<'db, L> {
 ///
 /// Querying the root of the trie will commit automatically.
 ///
-/// Dropping the instance may or may not commit depending on how it was constructed:
-/// - Instances created with [`TrieDBMutBuilder::build`] will commit on drop.
-/// - Instances created with [`TrieDBMutBuilder::build_base`] will not commit on drop.
+/// Dropping the instance may or may not commit depending on [Self::commit_on_drop] flag.
 ///
 /// # Example
 /// ```ignore
--- a/trie-db/src/triedbmut.rs
+++ b/trie-db/src/triedbmut.rs
@@ -22,7 +22,7 @@ use crate::{
 		NodeKey, NodeOwned, Value as EncodedValue, ValueOwned,
 	},
 	node_codec::NodeCodec,
-	rstd::{boxed::Box, convert::TryFrom, mem, ops::Index, result, vec::Vec, VecDeque},
+	rstd::{boxed::Box, convert::TryFrom, mem, ops, ops::Index, result, vec::Vec, VecDeque},
 	Bytes, CError, CachedValue, DBValue, Result, TrieAccess, TrieCache, TrieError, TrieHash,
 	TrieLayout, TrieMut, TrieRecorder,
 };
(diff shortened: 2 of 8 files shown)


trie-db/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "trie-db"
-version = "0.30.0"
+version = "0.31.0"
```
lexnv, line 3: nit: I generally prefer having a dedicated "chore" release PR with no/minimal changes, we'll also need to add this PR to the trie-db changelog 🤔

trie-db/src/triedbmut.rs
```diff
@@ -753,7 +776,31 @@ where
 	recorder: Option<core::cell::RefCell<&'a mut dyn TrieRecorder<TrieHash<L>>>>,
 }
 
-impl<'a, L> TrieDBMut<'a, L>
+/// Wrapper around `TrieDBMutBase` that commits changes on drop.
```
lexnv: nit: `[TrieDbMutBase]` for doc links? Similar below

trie-db/src/triedbmut.rs
```diff
@@ -2098,15 +2145,59 @@ where
...
+		self.0.commit();
 	}
 }
 
+/// Type alias for `Trie` implementation using a generic `HashDB` backing database.
+///
+/// Can be used as a `TrieMut` trait object. `db()` can be used to get the backing database
+/// object.
+///
+/// Note that no changes are committed to the database until `commit` is called.
```
lexnv: nit: Tiny bit confusing to me, maybe merge these 2 lines into something similar to "until `commit` is called, the root is queried or the object is dropped"?

trie-db/src/triedbmut.rs
```diff
@@ -2098,15 +2145,59 @@ where
...
+///
+/// Querying the root or dropping the instance will commit automatically.
```
lexnv: Is the `fn root()` ++ `fn drop()` a noop in case no keys were added in-between?

In other words:

```rust
let mut trie = CommitOnDrop..;

trie.insert(key, val);

// T0. At this point we perform a `commit` under the hood?
let root = trie.root();

// T1. Do we redo the `commit` at this point although the trie is already commited and it suffered no changes?
drop(trie);

**Review by lexnv — APPROVED** [review]
LGTM! I think we can add the release to a dedicated PR and tiny question around the `root ++ commit` op 🙏

File: trie-db/Cargo.toml (hunk 1 above)
Comment (michalkucharczyk) [reply trie-db/Cargo.toml:3]: well, just copied the "modus operandi" from some other [commit](https://github.com/paritytech/trie/commit/0f30619e2132d521bd6e97e5a13a5938981d1d5d) in this repo :), which makes a change and bumps version.

Comment (lexnv) [reply trie-db/Cargo.toml:3]: I think that works too, proly need a few more bumps? 🤔

File: trie-db/src/triedbmut.rs (hunk 4 above)
Comment (michalkucharczyk) [reply trie-db/src/triedbmut.rs]: `drop` always calls `commit`, exactly as it was before this PR.

File: trie-db/src/triedbmut.rs (hunk 2 above)
Comment (michalkucharczyk) [reply trie-db/src/triedbmut.rs]: a5c135d

File: trie-db/src/triedbmut.rs (hunk 3 above)
Comment (michalkucharczyk) [reply trie-db/src/triedbmut.rs]: reworked in: a5c135d
let me know if that sounds better?

File: trie-db/src/triedbmut.rs (hunk 4 above)
Comment (michalkucharczyk) [reply trie-db/src/triedbmut.rs]: and, yeah commit should be no-op at least from what I understand.

File: trie-db/src/triedbmut.rs [hunk 5]
```diff
@@ -2098,15 +2145,60 @@ where
...
+/// Can be used as a [`TrieMut`] trait object. `db()` can be used to get the backing database
+/// object.
+///
+/// Querying the root or dropping the instance will commit changes.
+///
+/// Note that no changes are committed to the database until `commit` is called explicitly or
+/// implicitly.
+///
+/// Refer to [`TrieDBMutBase`] for `Trie` implementation that does not perform `commit` on `drop`.
+pub type TrieDBMut<'a, L> = CommitOnDrop<'a, L>;
```
alexggh: Did you point polkadot-sdk to this branch an confirm everything still properly compiles, to make sure no unpleasant surprises appear after merge.

trie-db/src/triedbmut.rs
```diff
@@ -753,7 +776,31 @@ where
 	recorder: Option<core::cell::RefCell<&'a mut dyn TrieRecorder<TrieHash<L>>>>,
 }
 
-impl<'a, L> TrieDBMut<'a, L>
+/// Wrapper around [`TrieDBMutBase`] that commits changes on drop.
+pub struct CommitOnDrop<'a, L: TrieLayout>(TrieDBMutBase<'a, L>);
```
alexggh: The naming here could be better CommitOnDrop doesn't tell you much this is still a TrieDbMutWithCommitOnDrop.

trie-db/src/triedbmut.rs
```diff
@@ -2098,15 +2145,60 @@ where
...
+	}
+}
+
+impl<'a, L> Drop for CommitOnDrop<'a, L>
 where
 	L: TrieLayout,
 {
 	fn drop(&mut self) {
-		self.commit();
+		self.0.commit();
```
alexggh: Personally, I would've just went with a flag here on TrieDbMut, commit_on_drop, 

if commit_on_drop self.0.commit(),

trie-db/src/triedbmut.rs
michalkucharczyk: Yes. But we still need to release and bump version in polkadot-sdk, so we have time to handle surprises.

trie-db/src/triedbmut.rs
michalkucharczyk: Basically this one one-liner type is self-explanatory, but, yeah I can make it longer if you wish so. There is also legacy alias which is supposed to be used.

trie-db/src/triedbmut.rs
michalkucharczyk: Yes, that's also a way to go. Somehow I concluded that typed approach will be more explicit.

Still can change this, any other opinions on this? (@lexnv @skunert)

skunert: I think both are fine, but I tend to lean towards flag approach. Simply because it should be less code to digest and inspect. And as reader a `commit_on_drop` flag is kind of easy to understand. Ofc default should keep the old behaviour.

michalkucharczyk: > Personally, I would've just went with a flag here on TrieDbMut, commit_on_drop,

done here: 3eaef2b

alexggh (approved):

lexnv (approved):
Looks good! 🙏

skunert (approved):

trie-db/src/triedbmut.rs
```diff
@@ -695,7 +696,24 @@ impl<'db, L: TrieLayout> TrieDBMutBuilder<'db, L> {
...
+	/// Disable automatic commit on drop.
+	///
+	/// By default, [`TrieDBMut`] automatically commits changes when dropped. Calling this method
+	/// disables that behavior, requiring explicit calls to [`TrieDBMut::commit`] to persist
+	/// changes to the database.
+	///
+	/// This is useful when you want fine-grained control over when changes are committed, or when
+	/// you want to avoid the performance cost of committing if you're just doing temporary
+	/// operations.
+	pub fn without_commit_on_drop(mut self) -> Self {
```
bkchr:
```suggestion
	pub fn disable_commit_on_drop(mut self) -> Self {
```

trie-db/src/triedbmut.rs
```diff
@@ -707,17 +725,21 @@ impl<'db, L: TrieLayout> TrieDBMutBuilder<'db, L> {
...
-/// Use it as a `TrieMut` trait object. You can use `db()` to get the backing database object.
+/// Use it as a [`TrieMut`] trait object. You can use `db()` to get the backing database object.
 /// Note that changes are not committed to the database until `commit` is called.
 ///
-/// Querying the root or dropping the trie will commit automatically.
+/// Querying the root of the trie will commit automatically.
 ///
+/// Dropping the instance may or may not commit depending on how it was constructed:
+/// - Instances created with [`TrieDBMutBuilder::build`] will commit on drop.
+/// - Instances created with [`TrieDBMutBuilder::build_base`] will not commit on drop.
```
bkchr: Stale comment?

bkchr (approved):
Some nitpicks. Otherwise should be fine as it doesn't change the default behavior.

trie-db/Cargo.toml
michalkucharczyk: @lexnv 

Yes, I can do it. Not really needed as they are mainly tests/benchmarks, but I will bump and publish these crates.
See: c682c3a

trie-db/src/triedbmut.rs
michalkucharczyk: done: d6e54c5

trie-db/src/triedbmut.rs
michalkucharczyk: renamed in: d6e54c5

michalkucharczyk:
Uploaded crates:
```
Uploaded trie-db-test v0.31.1 to registry `crates-io`
Uploaded trie-bench v0.42.1 to registry `crates-io`
Uploaded reference-trie v0.29.2 to registry `crates-io`
Uploaded trie-db v0.31.0 to registry `crates-io`
```

New tags:
```
 * [new tag]         reference-trie-v0.29.2 -> reference-trie-v0.29.2
 * [new tag]         trie-bench-v0.42.1 -> trie-bench-v0.42.1
 * [new tag]         trie-db-test-v0.31.1 -> trie-db-test-v0.31.1
 * [new tag]         trie-eip1186-v0.5.1 -> trie-eip1186-v0.5.1
 * [new tag]         trie-eip1186-test-v0.7.1 -> trie-eip1186-test-v0.7.1
 * [new tag]         trie-db-v0.31.0 -> trie-db-v0.31.0
```

PR #227, opened 2025-10-02
fix decode compact prefix db

turuslan:
- partial prefix for attached value was not appended
- <strike>add support for child trie root prefix</strike> (use prefixed db wrapper)
- also [backported](https://github.com/turuslan/paritytech-trie/tree/fix/decode_compact_prefix_db_0_30_0) for https://github.com/paritytech/polkadot-sdk/pull/9247

### DIFF
--- a/trie-db/test/src/trie_codec.rs
+++ b/trie-db/test/src/trie_codec.rs
@@ -122,7 +122,13 @@ fn trie_compact_encoding_works_internal<T: TrieLayout>() {
 
 	encoded.push(Vec::new()); // Add an extra item to ensure it is not read.
 	test_decode_compact::<T>(MemoryDB::<T>::default(), &encoded, &items, root, encoded.len() - 1);
-	test_decode_compact::<T>(PrefixedMemoryDB::<T>::default(), &encoded, &items, root, encoded.len() - 1);
+	test_decode_compact::<T>(
+		PrefixedMemoryDB::<T>::default(),
+		&encoded,
+		&items,
+		root,
+		encoded.len() - 1,
+	);
 }
 
 test_layouts!(
--- a/trie-db/test/src/trie_codec.rs
+++ b/trie-db/test/src/trie_codec.rs
@@ -25,6 +25,12 @@ type MemoryDB<T> = memory_db::MemoryDB<
 	DBValue,
 >;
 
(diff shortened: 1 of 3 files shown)


bkchr (commented):
Please provide some test for these changes.

trie-db/test/src/trie_codec.rs
```diff
@@ -76,7 +82,7 @@ fn test_decode_compact<L: TrieLayout>(
 	expected_used: usize,
 ) {
 	// Reconstruct the partial DB from the compact encoding.
-	let mut db = MemoryDB::<L>::default();
+	let mut db = PrefixedMemoryDB::<L>::default();
```
bkchr: You can not just change an existing test if you introduce new code. We for sure want to ensure that both still work.

Also you don't use your newly introduced function.

turuslan: I made new function `decode_compact_from_iter_with_prefix` to avoid breaking interface (new `root_prefix` arg).
Old `decode_compact_from_iter` just calls `decode_compact_from_iter_with_prefix` with empty prefix.

This test with `PrefixedMemoryDB` panics on master, so this branch fixes existing functionality.

trie-db/test/src/trie_codec.rs
```diff
@@ -115,7 +121,8 @@ fn trie_compact_encoding_works_internal<T: TrieLayout>() {
 	);
 
 	encoded.push(Vec::new()); // Add an extra item to ensure it is not read.
-	test_decode_compact::<T>(&encoded, items, root, encoded.len() - 1);
+	test_decode_compact::<T>(MemoryDB::<T>::default(), &encoded, &items, root, encoded.len() - 1);
+	test_decode_compact::<T>(PrefixedMemoryDB::<T>::default(), &encoded, &items, root, encoded.len() - 1);
```
bkchr: You are still not testing the function that you introduced.

turuslan: Thanks, removed new function.
Running this test on master (before this PR) will panic, keeping it as regression test.

bkchr: Okay perfect.

Harrm (approved):

eskimor (approved):

bkchr (approved):

PR #233, opened 2026-07-07
CompactProof: Deduplicate values

skunert:
## Summary

`encode_compact` re-emits a shared detached value once per referencing node, causing proof size to grow with reference count instead of value count. Adds opt-in `encode_compact_skip_duplicate_values` to dedupe on encode. `decode_compact_from_iter` (and `decode_compact`) now always handle both old-style and deduplicated proofs correctly. Also fixes an off-by-one in `decode_compact_from_iter`'s returned item count when the last node has an attached value.

## Compatibility

- `encode_compact` / `decode_compact_from_iter` behavior unchanged — new functions are additive.
- Deduplicated proofs decode correctly with this crate's decoder, and with an unpatched **old** decoder into a hash-keyed `MemoryDB` (verified against 0.31.0) — but **not** into a `PrefixedKey` DB, which needs the value re-inserted per prefix.
- Old decoder + hash-keyed DB: refcount for the shared value is undercounted (1 instead of N). Reads still work (`get`/`contains` only check `rc > 0`); This matters if you later do balanced insert/remove against that decoded DB.
- `decode_compact`'s item count is now correct when the last node has an attached value — previously it was off by one in that case.


trie-db/src/trie_codec.rs
```diff
@@ -24,15 +24,31 @@
 //! hash references to nodes not in the partial trie are left intact. The compact encoding can be
 //! expected to save roughly (n - 1) hashes in size where n is the number of nodes in the partial
 //! trie.
+//!
+//! For layouts storing large values as separate, hash-addressed value nodes (see
+//! `TrieLayout::MAX_INLINE_VALUE`), a value node in the partial trie is "detached": the
```
bkchr:
```suggestion
//! [`TrieLayout::MAX_INLINE_VALUE`]), a value node in the partial trie is "detached": the
```

trie-db/src/trie_codec.rs
```diff
@@ -214,9 +238,39 @@ fn detached_value<L: TrieLayout>(
...
+/// Variant of [`encode_compact`] that emits each distinct detached value only once; later
+/// referencing nodes are emitted unmodified, referencing the value by hash.
+///
+/// `seen_value_hashes` collects the emitted value hashes. Pass `&mut Default::default()` for a
+/// standalone encoding, or thread the same set across concatenated encodings (the matching decode
+/// calls must then thread a known-values map through
+/// [`decode_compact_from_iter_with_known_values`]).
+pub fn encode_compact_skip_duplicate_values<L>(
+	db: &TrieDB<L>,
+	seen_value_hashes: &mut BTreeSet<Vec<u8>>,
```
bkchr: Does someone requires this after the encoding?

trie-db/src/trie_codec.rs
```diff
@@ -214,9 +238,39 @@ fn detached_value<L: TrieLayout>(
...
+	encode_compact_inner(db, None)
+}
+
+/// Variant of [`encode_compact`] that emits each distinct detached value only once; later
+/// referencing nodes are emitted unmodified, referencing the value by hash.
+///
+/// `seen_value_hashes` collects the emitted value hashes. Pass `&mut Default::default()` for a
+/// standalone encoding, or thread the same set across concatenated encodings (the matching decode
+/// calls must then thread a known-values map through
+/// [`decode_compact_from_iter_with_known_values`]).
```
bkchr:
```suggestion
/// `seen_value_hashes` collects the emitted value hashes.
```

Ty claude :P

bkchr (commented):
You can do the de-duplication on the node level and not just on the `value` level.

trie-db/src/trie_codec.rs
dmitry-markin: What is the use case for "concatenated encodings"? Do we actually need to share the set across calls, or can get rid of this argument?

trie-db/src/trie_codec.rs
skunert: We could use it for example [here](https://github.com/paritytech/polkadot-sdk/blob/d988144f162df86614159e932493bfa23b7ab5b3/substrate/primitives/trie/src/trie_codec.rs#L159) to deduplicate across child and main trie. Whether its really necessary to deduplicate across main and child is not super clear to me, but I can see use cases for this.

skunert:
> You can do the de-duplication on the node level and not just on the `value` level.

We now also deduplicate nodes

trie-db/src/iterator.rs
```diff
@@ -125,6 +125,19 @@ impl<L: TrieLayout> TrieDBRawIterator<L> {
 			.push(Crumb { hash: node_hash, status: Status::Entering, node: Arc::new(node) });
 	}
 
+	/// Skip the descendants of the node most recently yielded by `next_raw_item`: iteration
+	/// continues with the node's next sibling (or an ancestor's).
+	///
+	/// Must only be called directly after `next_raw_item(_, true)` yielded a node. At that point
+	/// the node's own partial and child index have not been pushed to `key_nibbles` yet (that
+	/// happens when its `Status::At` is processed), so the crumb can jump straight to
+	/// `Status::AftExiting` — popping it and stepping the parent — with nothing to unwind.
```
bkchr:
```suggestion
```

I don't want to read implementation details in the docs. Instead of having this written as docs, the code should  ensure that it is at a stage where the skipping works. And maybe also return a `bool` if this worked. Yeah for sure not important for  what you are doing here, but for our future.

trie-db/src/trie_codec.rs
```diff
@@ -241,8 +299,21 @@ where
 			Ok((prefix, node_hash, node)) => {
 				// Skip inline nodes, as they cannot contain hash references to other nodes by
 				// assumption.
-				if node_hash.is_none() {
-					continue
+				let node_hash = match node_hash {
+					Some(node_hash) => node_hash,
+					None => continue,
+				};
```
bkchr:
```suggestion
				let Some(node_hash) = node_hash else 
					continue
				};
```

trie-db/src/trie_codec.rs
```diff
@@ -241,8 +299,21 @@ where
...
+
+				if let Some(seen_hashes) = seen_hashes.as_deref_mut() {
+					// A subtree whose root was already emitted is skipped entirely; the parent's
+					// `omit_children` bit stays unset, keeping a plain hash reference. The root is
+					// never skipped, so each encoding stays individually decodable when
+					// `seen_hashes` is threaded across successive encodings.
+					if !stack.is_empty() && seen_hashes.contains(node_hash.as_ref()) {
```
bkchr: I don't get why you check here if `stack` is not empty?

trie-db/src/trie_codec.rs
```diff
@@ -431,6 +527,128 @@ impl<'a, C: NodeCodec> DecoderStackEntry<'a, C> {
...
+	DB: HashDB<L::Hash, DBValue>,
+{
+	let partial_len = match &entry.node {
+		Node::Extension(partial, _) | Node::NibbledBranch(partial, _, _) => {
+			prefix.append_partial(partial.right());
+			partial.len()
+		},
+		_ => 0,
+	};
+	for index in 0..entry.children.len() {
```
bkchr: Why are you not directly iterating the `entry.children`?

trie-db/src/trie_codec.rs
```diff
@@ -431,6 +527,128 @@ impl<'a, C: NodeCodec> DecoderStackEntry<'a, C> {
...
+		if entry.hash_ref_children & (1u16 << index) == 0 {
+			continue
+		}
```
bkchr:
```suggestion
```

This is the exact same check as below?

trie-db/src/trie_codec.rs
```diff
@@ -318,6 +405,9 @@ struct DecoderStackEntry<'a, C: NodeCodec> {
 	child_index: usize,
 	/// The reconstructed child references.
 	children: Vec<Option<ChildReference<C::HashOut>>>,
+	/// Bit mask of children kept as plain hash references, which may point at subtrees
+	/// deduplicated into an earlier occurrence (see [`encode_compact_skip_duplicates`]).
+	hash_ref_children: u16,
```
bkchr:
```suggestion
	has_ref_children: bool,
```

And then just set it to `true` if there is one.

trie-db/src/trie_codec.rs
```diff
@@ -341,6 +431,9 @@ impl<'a, C: NodeCodec> DecoderStackEntry<'a, C> {
 						let child_ref = child.try_into().map_err(|hash| {
 							Box::new(TrieError::InvalidHash(C::HashOut::default(), hash))
 						})?;
+						if let ChildReference::Hash(_) = child_ref {
```
bkchr:
```suggestion
						if child_ref.is_hash() {
```

Maybe?

trie-db/src/trie_codec.rs
```diff
@@ -525,16 +778,47 @@ where
...
+			known_items.insert(node_hash.as_ref().to_vec(), node_data);
+
 			if let Some(entry) = stack.pop() {
 				last_entry = entry;
 				last_entry.pop_from_prefix(&mut prefix);
 				last_entry.children[last_entry.child_index] = Some(ChildReference::Hash(node_hash));
 				last_entry.child_index += 1;
 			} else {
-				return Ok((node_hash, i + 1))
+				return Ok((node_hash, used))
```
bkchr:
```suggestion
				return Ok((node_hash, i + 1 + attached_node))
```

Also it would be nice if we could explain in the docs what the value is :P

trie-db/src/trie_codec.rs
```diff
@@ -525,16 +778,47 @@ where
 				prefix.drop_lasts(partial_prefix_len);
 				hash
 			});
+			if hash.is_none() {
+				// The node's detached value may have been deduplicated into an earlier occurrence
+				// (see `encode_compact_skip_duplicates`); re-insert it here too so `db` matches an
+				// un-deduplicated encoding. A miss means the value is not in the encoding, which is
```
bkchr: If the hash is `none` and the the value is missing, it means the db is missing values and we should return an error?

trie-db/src/trie_codec.rs
```diff
@@ -459,6 +677,31 @@ pub fn decode_compact_from_iter<'a, L, DB, I>(
...
+/// database as an un-deduplicated encoding. Threading across calls is only needed for encodings
+/// deduplicated with a shared `seen_hashes` set.
+///
+/// Decoding work is proportional to the *un-deduplicated* encoding: re-inserting subtrees at every
+/// occurrence can touch far more positions than there are items in `encoded`, so callers decoding
+/// untrusted input should bound the logical size, not the encoded size.
+pub fn decode_compact_from_iter_with_known_items<'a, L, DB, I>(
+	db: &mut DB,
+	encoded: I,
+	known_items: &mut BTreeMap<Vec<u8>, DBValue>,
```
bkchr: Do we actually need to supply this?

trie-db/src/trie_codec.rs
```diff
@@ -459,6 +677,31 @@ pub fn decode_compact_from_iter<'a, L, DB, I>(
 	db: &mut DB,
 	encoded: I,
 ) -> Result<(TrieHash<L>, usize), TrieHash<L>, CError<L>>
+where
+	L: TrieLayout,
+	DB: HashDB<L::Hash, DBValue>,
+	I: IntoIterator<Item = &'a [u8]>,
+{
+	decode_compact_from_iter_with_known_items::<L, DB, I>(db, encoded, &mut BTreeMap::new())
```
bkchr: This method is also doing the same reconstruction of  the database with the same counts. So, there is no real difference between these two?

skunert: The difference is just backwards compat. The previous method existed before and does not break anything. `with_known_items` is only needed if you want to pass through the `known_items`, which we could do in the scenario outlined [here](https://github.com/paritytech/trie/pull/233/changes/75728c18e769da384d4bd284a9b9ba215d023c79#r3569900092).

trie-db/src/trie_codec.rs
skunert: https://github.com/paritytech/trie/pull/233/changes#r3581312671

trie-db/src/trie_codec.rs
skunert: This is needed to decide whether we are at the root of a child trie for example. If the root would be in seen_hashes, the encoding for the child trie would be empty.

trie-db/src/trie_codec.rs
skunert: This is a bitmask that marks which children came on the wire as hash. We insert the deduplicated subtrees towards the end and iterate all the children. But at that point it is not possible to distinguish between items that came as detached value and items that where always Hash.

The ones that came as detached value where already inserted into the db, so we should not insert them again to keep the ref count in the db correct.

trie-db/src/trie_codec.rs
skunert: Because we use it to access the bitmask

trie-db/src/trie_codec.rs
skunert: See comment [here](https://github.com/paritytech/trie/pull/233/changes#r3582402655)

trie-db/src/trie_codec.rs
skunert: What does value missing mean for you here?

If the hash is none that means that there was no detached node in the compact encoding. So we just have the hash.

If that hash belongs to a node that was deduplicated and we have seen it earlier, we need to reinsert to maintain refcount.

if the hash belongs to a node that we have not yet seen, that means the node is just not referenced in the proof, which should also be fine.

skunert:
@cheme Would be super nice if you could support us with a review here :)

cheme:
> @cheme Would be super nice if you could support us with a review here :)

Sure, will look at it (may not be able to do tomorrow, but will try to do before next week).

trie-db/test/src/trie_codec.rs
```diff
@@ -152,6 +154,395 @@ fn trie_decoding_fails_with_incomplete_database_internal<T: TrieLayout>() {
 	}
 }
 
+/// A value above every tested layout threshold, stored as a shared, hash-addressed value node.
+const SHARED_VALUE: &[u8] = &[4; 32];
```
cheme: Would just switch to 33 to have more case with  value nodes (but no case with inline value).

trie-db/src/trie_codec.rs
```diff
@@ -187,24 +201,32 @@ impl<C: NodeCodec> EncoderStackEntry<C> {
...
-			}
-		},
+	let hash_plan = match value {
+		ValuePlan::Node(hash_plan) => hash_plan,
 		_ => return None,
+	};
+	let value_hash = &node_data[hash_plan.clone()];
+	if let Some(seen_hashes) = &seen_hashes {
+		if seen_hashes.contains(value_hash) {
+			return None
```
cheme, line 224:
```suggestion
			// Do not reinsert duplicated value node
			return None
```
Or any comment, just to show this is the line doing the work

trie-db/src/trie_codec.rs
```diff
@@ -241,8 +299,18 @@ where
 			Ok((prefix, node_hash, node)) => {
 				// Skip inline nodes, as they cannot contain hash references to other nodes by
 				// assumption.
-				if node_hash.is_none() {
-					continue
+				let Some(node_hash) = node_hash else { continue };
+
+				if let Some(seen_hashes) = seen_hashes.as_deref_mut() {
+					// A subtree whose root was already emitted is skipped entirely; the parent's
```
cheme, line 315:
```suggestion
					// Avoid skipping entirely a subtree whose root was already emitted; the parent's
```

trie-db/src/trie_codec.rs
cheme: I am not getting that, we skip only dettached value? not duplicated trie node hash (could make sense though)? so would only need to look at the single value hash?

trie-db/src/trie_codec.rs
```diff
@@ -214,9 +236,50 @@ fn detached_value<L: TrieLayout>(
...
+/// node set: a skipped subtree is reconstructable only if everything below it was emitted when
+/// its root was first seen. Encoding from per-proof recorded sets whose coverage of a shared
+/// node diverges silently drops the divergent nodes and produces unverifiable proofs.
+///
+/// Any decoder reconstructs a readable hash-keyed database from the result. Reconstructing the
+/// exact database of an un-deduplicated encoding (every position populated, matching reference
+/// counts) requires a re-inserting decoder, i.e. [`decode_compact_from_iter_with_known_items`].
+///
+/// Assumes occurrences of an item are interchangeable, as they are when `db` is hash-keyed. A
+/// prefixed `db` populating a duplicated subtree only below a later occurrence would drop it.
```
cheme: I am not sure I understand the sentence about Prefixed Db.

From what I remember a prefixed memory db should only be used with a trie to produce/record the payload to send to rocksdb. So generally I would think things would get easier if we avoid using it at all when doing a compact proof.
I may have missed a case (I remember seeing some fix passing, maybe about it), but right now I could not figure it to be usefull.

It seems in the PR, managing prefixed db is the main source of complexity. For instance, for subtree, if not requiring prefixed db, one can just record the first subtree and skip the other ones.

Similarily, Iirc, proofs do not need to have proper rc (rc of memory db being usefull only for the product/record the payload to send to paritydb). Execution against a state machine never change the original state (delta of a removed value is simply stored in a keyvalue change map, so the backend would not change until receiving the payload).

trie-db/src/trie_codec.rs
```diff
@@ -241,8 +304,21 @@ where
...
+				let Some(node_hash) = node_hash else { continue };
+
+				if let Some(seen_hashes) = seen_hashes.as_deref_mut() {
+					let is_root = stack.is_empty();
+					// A subtree whose root was already emitted is skipped entirely; the parent's
+					// `omit_children` bit stays unset, keeping a plain hash reference. The root is
+					// never skipped, so each encoding stays individually decodable when
+					// `seen_hashes` is threaded across successive encodings. Sound only under the
+					// fixed-backing-set precondition (see `encode_compact_skip_duplicates`): the
+					// subtree below a seen hash must not have grown since it was emitted.
```
cheme, line 320: I am not too sure this is needed, generally proof is a record of state accesses, so would only need to be an unprefixed memorydb with any rc being fine.

At the trie level it may make sense still, but subtree/childtree is a substrate concept, it is odd to see it in the trie crate.

trie-db/src/trie_codec.rs
```diff
@@ -431,6 +533,128 @@ impl<'a, C: NodeCodec> DecoderStackEntry<'a, C> {
 	}
 }
 
+/// Re-insert the deduplicated subtrees that `entry`'s node references through plain hash-reference
+/// children (see [`encode_compact_skip_duplicates`]), replaying what an un-deduplicated encoding
+/// would have inserted below this node. Hashes missing from `known_items` reference items outside
+/// the encoding and are skipped, like the holes an un-deduplicated encoding would leave.
+///
+/// `prefix` is `entry`'s node prefix; it is restored before returning.
```
cheme: IIUC the point of this function is to reinsert a subtree/childtree , which somehow would make sense to me to deduplicate over a different prefix, or account for proper Rc. (but as mentioned before I am unsure it make sense for a proof).

if needed, I would certainly feel like this should be part of substrate (yet I remember in the past having to put some substrate related design in the trie_codec, so there may be a reason here, but I don t see it immediatly).

trie-db/src/trie_codec.rs
```diff
@@ -525,16 +787,47 @@ where
 				prefix.drop_lasts(partial_prefix_len);
 				hash
 			});
+			if hash.is_none() {
+				// The node's detached value may have been deduplicated into an earlier occurrence
+				// (see `encode_compact_skip_duplicates`); re-insert it here too so `db` matches an
+				// un-deduplicated encoding. A miss means the value is not in the encoding, which is
+				// legal.
```
cheme: I see the point of reinserting, yet we are no sure the value was actually accessed here.

I would just not reinsert, as rc does not matter to me. If rc matters (not sure why), then we can insert indeed, but since we don’t know if it was skipped due to duplication or value unaccessed, we may obtain a wrong rc.

One solution might be to simply have another placeholder value for duplicated value, but then we are not really compatible with old proofs.

I wonder if we shall force proof to run on a unprefixed memorydb that do not allow insert and removal, so Rc or prefix would not matter.

cheme (commented):
Using seen_hashes make sense to me, I think it is good (should even be extended node hashes).
What I am not too sure, is why we try to use prefixedmemorydb or have proper rc count.

lexnv (approved):
LGTM! Had a look over comments and addresses them in https://github.com/paritytech/trie/pull/234

Maybe one extra thing before merging would be https://github.com/paritytech/trie/pull/234/changes/855ad48d1d1da350b6d13f14a43233f0690d66ee

skunert:
@cheme @bkchr Sorry for the delay (was off for some time), did another pass and removed the RC reconstruction and subtree reinsertion during decoding. Substrate does not need it at all, and it was complicating the code.

The reinsertion had another problem, a deduplicated proof emits each shared subtree once, but the reinserting decoder replayed it at every referencing position. Nested duplication makes that unfold exponentially costing exponential decode time. To fix this we would have needed to introduce some kind of decoding budget on the decoder side. Not worth it IMO.

And if needed, it is still possible to arrive at a correct PrefixedDB from the HashDB.

trie-db/src/trie_codec.rs
```diff
@@ -534,7 +631,7 @@ where
 				last_entry.children[last_entry.child_index] = Some(ChildReference::Hash(node_hash));
 				last_entry.child_index += 1;
 			} else {
-				return Ok((node_hash, i + 1))
+				return Ok((node_hash, i + 1 + attached_node))
```
cheme, line 656: I guess this change means major version update.

cheme (approved):
Looks good to me.
Note that I did not go through all testing and fuzzing code path, from a quick glance, maybe there is a bit of code that is not so needed anymore (regarding latest simplification).

lexnv (approved):
Had a brief look, this also incorporates the hashing optimization via `Copy` (ie `TrieHash<L>`)

One tiny thing that can be a followup would be to turn Clippy/Fmt unused/wrongly formatting warnings into errors at compile time such the CI would fail instead of the bot leaving comments on the PR, worth considering later 🙏

trie-db/src/trie_codec.rs
```diff
@@ -187,24 +201,35 @@ impl<C: NodeCodec> EncoderStackEntry<C> {
...
+	if let (Some(seen_hashes), Some(key)) = (&seen_hashes, &dedup_key) {
+		// Already emitted once: keep the plain hash reference instead of detaching again.
+		if seen_hashes.contains(key) {
+			return None
+		}
+	}
+	let fetched = match TrieDBRawIterator::fetch_value(db, value_hash, node_prefix) {
+		Ok(value) => value,
+		Err(_) => return None,
+	};
```
bkchr:
```suggestion
	let fetched = TrieDBRawIterator::fetch_value(db, value_hash, node_prefix).ok()?;
```

bkchr (approved):

PR #234, opened 2026-07-31
trie-db: Avoid cloning the hashes on the hot path

lexnv:
This PR addresses the comments from https://github.com/paritytech/trie/pull/233 and optimizes the `seen_hashes` to take an `TrieHash` instead of allocating a vec.


skunert:
Has been separately integrated in #233

PR #239, opened 2026-09-16
ci: publish crates through crates_publish_automation

gab8i:
Moves publishing from manual `cargo publish` to paritytech/crates_publish_automation. Publishing a GitHub Release now packages the publishable crates, uploads them as an artifact, and dispatches the central publisher, which publishes whatever crates.io does not have yet.


test-support/keccak-hasher/Cargo.toml
```diff
@@ -1,5 +1,8 @@
 [package]
 name = "keccak-hasher"
+# NOTE: do not bump this version. `parity-crate-owner` does not own this crate
```
skunert: I think we can remove these comments. We will only merge here once parity-crate-owner owns these crates right?

skunert (approved):

test-support/keccak-hasher/Cargo.toml
gab8i: Yes, you're right, they were part of the draft but given the order of things which are expected to happen they can be removed!

